PR #15, opened 2021-04-06
fix typos in join

kiyonlin:


muesli:
Thank you for all these cleanup PRs, @kiyonlin!

PR #49, opened 2021-11-22
Fix JoinVertical behavior for non-edge non-center alignments

ryantriangles:
Currently, JoinVertical displays like this:

```
JoinVertical(0.00, "A", "BBBB")
A   
BBBB


JoinVertical(0.25, "A", "BBBB")
  A 
BBBB


JoinVertical(0.75, "A", "BBBB")
 A  
BBBB


JoinVertical(1.00, "A", "BBBB")
   A
BBBB
```

It seems more natural that 0.25 would be 25% of the way from 0 to 1, and 0.75 would be 75% of the way from 0 to 1. This pull request switches to that behavior, and adds a regression test for the change.

### DIFF
--- a/join.go
+++ b/join.go
@@ -157,8 +157,8 @@ func JoinVertical(pos Position, strs ...string) string {
 				}
 
 				split := int(math.Round(float64(w) * pos.value()))
-				left := w - split
-				right := w - left
+				right := w - split
+				left := w - right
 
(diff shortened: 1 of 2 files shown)


meowgorithm:
Thanks so much for this one: it's been a longstanding one we've been meaning to fix.

PR #51, opened 2021-12-10
Fix README.md typo

stefanvanburen:
Looks like an extra word got in there? 😄


meowgorithm:
Alas, indeed. Thanks for the correction!

Issue #52, opened 2021-12-24
Allow disabling alignment during "Render"


Maverobot:
It seems that currently strings with multiple lines are always aligned during rendering. See https://github.com/charmbracelet/lipgloss/blob/4f0d44/style.go#L333-L346

I would like to disable the alignment in my use case. It would make sense to let the user choose if an alignment is desired. What do you think?

PR #53, opened 2021-12-26
Allow disabling alignment when rendering multiple lines

Maverobot:
This is a possible solution for this reported issue of mine - https://github.com/charmbracelet/lipgloss/issues/52


meowgorithm:
@Maverobot Sorry for the slow reply to this one (been a busy start to the year). Can you provide some sample code to help us understand the use case?

PR #54, opened 2022-01-07
Add separate sets of hard- and soft-enforced linters

muesli:
Soft-enforced linters will only annotate code changes, hard-enforced linters actually make the workflow fail if they encounter an issue.

You can run the soft linters manually on your machine:

```bash
golangci-lint run --config .golangci-soft.yml
```


meowgorithm:
Thanks, @muesli!

PR #60, opened 2022-01-31
Use termenv to enable Windows virtual terminal processing

muesli:
Keeps it in a central place at the bottom of the stack. Should eventually migrate the other apps using this little code snippet as well.


meowgorithm (approved):

meowgorithm:
Good call. LGTM!

PR #66, opened 2022-02-11
Add Value method

76creates:
Simple method for checking the value of the style set with `SetString`. This would be useful when creating conditionals for passed style objects.

### DIFF
--- a/style.go
+++ b/style.go
@@ -97,6 +97,11 @@ func (s Style) SetString(str string) Style {
 	return s
 }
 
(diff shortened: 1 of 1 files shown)


muesli:
I know this isn't entirely consistent with the rest of the operations in `borders.go` and `get.go`, but I think this should just be called `Value` (no `Get` prefix), as there's no operation involved.

cc @meowgorithm

meowgorithm:
Agreed, `Value` is the way to go here.

muesli:
Thank you @76creates!

PR #69, opened 2022-02-21
don't overwrite background when inheriting

76creates:
It's contra intuitive that `backgroundKey` is the only key that is **always inherited**.

### DIFF
--- a/example/vet/main.go
+++ b/example/vet/main.go
@@ -1,1 +1,0 @@
-package vet
--- a/example/vet/main.go
+++ b/example/vet/main.go
@@ -1,0 +1,1 @@
+package vet
--- a/style.go
+++ b/style.go
@@ -137,6 +137,9 @@ func (s Style) Inherit(i Style) Style {
 			// Padding is not inherited
 			continue
 		case backgroundKey:
+			if _, exists := s.rules[k]; exists {
+				continue
+			}
 			s.rules[k] = v
 
 			// The margins also inherit the background color
--- a/style.go
+++ b/style.go
@@ -137,11 +137,6 @@ func (s Style) Inherit(i Style) Style {
 			// Padding is not inherited
 			continue
 		case backgroundKey:
-			if _, exists := s.rules[k]; exists {
-				continue
-			}
-			s.rules[k] = v
-
 			// The margins also inherit the background color
 			if !s.isSet(marginBackgroundKey) && !i.isSet(marginBackgroundKey) {
 				s.rules[marginBackgroundKey] = v


meowgorithm:
Thanks for this! Would you mind posting a tiny bit of sample code to help explain why this makes sense? I know it might seem trivial, but it'll help expedite this.

Thanks!

76creates:
> Thanks for this! Would you mind posting a tiny bit of sample code to help explain why this makes sense? I know it might seem trivial, but it'll help expedite this.
> 
> Thanks!

From the function point of view `Inherit`, and any other fn, should be as consistent as possible, meaning that we should minimise special cases like those with background and margin background keys as they might lead to difficult to debug situations. This is my motive for this PR.

From the user point of view, in my case that is, I was looking for a way to set a _default style_ for an array of Style objects, we can normally achieve that using `Copy` but in this case both the _inheritor_ and _origin_ objects mutate over time and are "joined" at one moment. I can go around this issue by checking if background is set, storing it if it is, then inheriting, and then overwriting the background, which is all a bit messy in my opinion.

meowgorithm:
Cool, so this makes sense. Basically if a background color is set on the inheritor it shouldn’t be overridden during inheritance. Clearly just something that was overlooked.

We’ll still need to vet this PR before merging, so if you have any sample code it’ll help expedite the process.

76creates:
> Cool, so this makes sense. Basically if a background color is set on the inheritor it shouldn’t be overridden during inheritance. Clearly just something that was overlooked.
> 
> We’ll still need to vet this PR before merging, so if you have any sample code it’ll help expedite the process.

https://gist.github.com/76creates/8f529f2b975f3cd6dee1d2a70260d53e

76creates:
for full example check out https://github.com/76creates/stickers
The problem appears [here](https://github.com/76creates/stickers/blob/f1a12c47674bacbed58542588b18af1d89d7dc1e/flexBoxCell.go#L95)

muesli:
> for full example check out https://github.com/76creates/stickers The problem appears [here](https://github.com/76creates/stickers/blob/f1a12c47674bacbed58542588b18af1d89d7dc1e/flexBoxCell.go#L95)

🤯 Nice work on stickers!

meowgorithm:
Thanks again for this. It's awesome seeing some good use of inheritance on [stickers](https://github.com/76creates/stickers).

76creates:
> Thanks again for this. It's awesome seeing some good use of inheritance on [stickers](https://github.com/76creates/stickers).

no, no, **thank you** folks for making such an awesome lib to begin with ⚡

PR #71, opened 2022-03-04
Clarify Inherit comment

Evertras:
The old comment implied that the argument would overwrite, but this is different from the actual behavior which is to NOT overwrite.  This updates the comment to reflect this behavior.


meowgorithm:
Thank you! Will merge as soon as CI wraps up.

PR #72, opened 2022-03-09
fix: double word wrap and ignoring style margins

aymanbagabas:
* fix word wrap ignore style left and right margins
* hard-wrap string when width is specified


meowgorithm:
Closing this one as it greatly alters paragraph rendering. We can revisit after https://github.com/muesli/reflow/pull/3 is merged.

PR #80, opened 2022-04-11
Add TestPlace for position.go

bashbunni:
## Changes:
- [x] add test for Place from position.go

## Upgrades:
- increase test coverage for position.go from 0% to 70.91%

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 ssh_example_ed25519*
+*.out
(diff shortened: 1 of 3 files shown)


maaslalani:
@bashbunni I think this is good to merge, one thing I would recommend is removing the `require` dependency and simply using go `==` for `string` comparison.

bashbunni:
I'll rebase and fix the conflicts, then see if it still works :joy:

bashbunni:
This should be covered in a test, but I'm going to close this one as I'd prefer to use a golden file for this

PR #91, opened 2022-06-29
feat: lipgloss renderer

aymanbagabas:
* Introduce lipgloss renderer to render styles and text
* Deprecate the stringer interface
* Drop TerminalColor `color.Color` conformance and reduce dependencies 
* Add `ANSIColor` as a syntactic sugar
* Update docs


renderer.go
```diff
@@ -0,0 +1,417 @@
...
+//
+// Available color profiles are:
+//
+// termenv.Ascii (no color, 1-bit)
+// termenv.ANSI (16 colors, 4-bit)
+// termenv.ANSI256 (256 colors, 8-bit)
+// termenv.TrueColor (16,777,216 colors, 24-bit)
+//
+// This function is thread-safe.
+func (r *Renderer) SetColorProfile(p termenv.Profile) {
```
meowgorithm, line 85: Does it make sense to have a `func WithColorProfile() RendererOption` in the same way we have `WithOutput` and `WithDarkBackground`?

muesli: I like that.

meowgorithm: Cool, just added.

meowgorithm (approved):

PR #97, opened 2022-07-29
Add new Border Title API 

eugener:
Adds new Border Title API:
- `func (s Style) BorderTitle(title string) Style`
- `func (s Style) BorderTitleAlignment(alignment Position) Style`
- `func (s Style) BorderTitleBackground(color TerminalColor) Style`
- `func (s Style) BorderTitleForeground(color TerminalColor) Style`
- `func (s Style) GetBorderTitle() string`
- `func (s Style) GetBorderTitleAlignment() Position`
- `func (s Style) GetBorderTitleBackground() TerminalColor` 
- `func (s Style) GetBorderTitleForeground() TerminalColor `

Updated example:
image


eugener:
Once the new API is agreed upon and stable, I will also update the documentation

maaslalani:
This looks really good @eugener, nice work!

eugener:
@maaslalani @meowgorithm Is there anything else needed to make this PR a part the next release?

meowgorithm:
Hey, @eugener! We just need a little time to give this a proper review. Thanks for your patience with this one; we'll definitely get to it.

eugener:
Sounds good! Let me know how I can help.

.gitignore
```diff
@@ -0,0 +1 @@
+**/.idea/
```
maaslalani, line 1: Hey there, do you mind deleting this file from the PR?

I highly recommend setting up a [global `.gitignore`](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/) to ignore these files from all git projects without needing to set it up in each project!

maaslalani:
Another question I had, was would we want to support Bold / Italic / Underline / Padding / Margin in the border title?

I almost wonder if the border title should take it's own style or take only a string which we expect the user to style before passing in?

.gitignore
eugener: Done.
Just have to note that, in my opinion, having `.gitignore` within your repo ensures consistency of commits without relying on environment external to the project.

eugener:
> Another question I had, was would we want to support Bold / Italic / Underline / Padding / Margin in the border title?
> 
> I almost wonder if the border title should take its own style or take only a string which we expect the user to style before passing in?
> 
> I.e:
> 
> ```go
> borderTitle := lipgloss.NewStyle().Bold(true).Underline(true).Render("Hello, Borders")
> lipgloss.NewStyle().BorderTitle(borderTitle).Border(lipgloss.DoubleBorder())
> ```

I think it is a great idea, but it complicates the API significantly. 
In my opinion, releasing the Border Title API in its current state solves the current need. 
Additional attributes can be added later if everyone agrees with their usefulness.

maaslalani:
I think there's two ways to do it, either we can get passed the styled text or take the `BorderTitleStyle` as an argument and then apply the `BorderTitle` as an [Inline Style](https://github.com/charmbracelet/lipgloss#enforcing-rules). I.e. remove the padding and margin.

The current API doesn't prevent people from passing in multi-line text (if I'm not mistaken), so I think it is reasonable to expect the user to pass in a non-multiline string / inline style. But, @meowgorithm may have more thoughts on this.

Personally, I think it would be reasonable to remove the `BorderTitleForeground` and `BorderTitleBackground` from the API and expect people to pass in a pre-styled one-line string which includes horizontal padding / fg / bg, etc...

What do you think @meowgorithm @eugener? I'm not too opinionated on this one.

maaslalani:
That being said, I am also fully onboard with the current PR as it is. I think it is currently in a shippable state.

meowgorithm:
We will need to offer at least the basic ANSI styling options for this (bold, underline, strikethrough, reverse, and so on) so it _probably_ makes sense to pass a style. And we do have mechanisms to keep help prevent people from breaking stuff (see [Enforcing Rules](https://github.com/charmbracelet/lipgloss#enforcing-rules) in the README).

So if we go down this route you'd either pass a styled string:

```go
borderEverything := lipgloss.NewStyle().
    Bold(true).
    Underline(true).
    Render("Hello, Borders")

style := lipgloss.NewStyle().
    BorderTitle(borderEverything).
    Border(lipgloss.DoubleBorder())
```

Or you'd separate concerns a bit more and set the style and the string separately:

```go
borderTitleStyle := lipgloss.NewStyle().
    Bold(true).
    Underline(true)

style := lipgloss.NewStyle().
    BorderTitleStyle(borderStyle).
    BorderTitle("Hello, Borders").
    Border(lipgloss.DoubleBorder())

```

eugener:
@maaslalani @meowgorithm Thank you both for very insightful suggestions!
Based on your suggestions, I redesigned the API by passing the `Style` instead of a text, which eliminated the need for all API methods except get/set BorderTitle. Currently, the title text is set using `style.SetString` method, which seems fine to me comparing to introducing a separate method for setting border title text on Style., but I'm for discussion on this topic. 
Example and docs are up-to-date also.

maaslalani:
I think using the `SetString` and the `HorizontalAlign` API is quite clever, you basically get the best of both worlds with the first (singular entry point) and second approach (flexibility and control of the Style).

meowgorithm:
This is great @eugener, though after thinking on it I would argue that it may make sense to split out the style from the value both to be explicit, and to separate style from data for re-use as I expect a common use case to be rendering several similar looking boxes with different titles. Consider the following:

```go
titleStyle := lipgloss.NewStyle().
    Bold(true).
    Reverse(true)

mrBoxStyle := lipgloss.NewStyle().
    BorderTitleStyle(borderStyle).
    BorderTitle("Mr. Box").
    Border(lipgloss.RoundedBorder())

mrsBoxStyle := mrBoxStyle.Copy().BorderTitle(“Mrs. Box”)
```

This is, of course, doable in the current implementation but a bit more roundabout:

```go
mrTitle := lipgloss.NewStyle().
    Bold(true).
    Reverse(true).
    SetString("Mr. Box")

mrsTitle := mrTitle.Copy().SetString(“Mrs. Box”)

mrBox := lipgloss.NewStyle().
    BorderTitle(mrBox).
    Border(lipgloss.RoundedBorder())

mrsBox := mrBox.Copy().BorderTitle(mrsTitle)
```

eugener:
@meowgorithm Agree! This makes border styles a lot more reusable. Will rework it as you suggest.

meowgorithm:
Sounds good. Thanks for all your hard work on this, @eugener!

eugener:
The API is now reworked, including docs, example and readme. Hopefully I did not miss anything.
@meowgorithm @maaslalani  Please review.

borders.go
```diff
@@ -296,6 +324,13 @@ func (s Style) applyBorder(str string) string {
 	return out.String()
 }
 
+func repeatStr(s string, count int) string {
```
maaslalani, line 327: This is sort of a nitpick so feel free to ignore, but instead of having a wrapper around `strings.Repeat` called `repeatStr` I would do something like this:

```go
strings.Repeat(border.Top, max(0, value))
```

Where `max` is:
```go
func max(a, b) int {
  if a > b {
    return a
  }
  return b
}
```

maaslalani (commented):
This is fantastic @eugener. Taking a look now!

borders.go
maaslalani: You can use the `max` defined here: https://github.com/charmbracelet/lipgloss/blob/0ce55509014b0cc476e197f22aa4d43add081a57/style.go#L460-L465

example/main.go
```diff
@@ -239,7 +246,7 @@ func main() {
 
 		dialog := lipgloss.Place(width, 9,
 			lipgloss.Center, lipgloss.Center,
-			dialogBoxStyle.Render(ui),
+			dialogBoxStyle.Copy().BorderTitle(" Question ").Render(ui),
```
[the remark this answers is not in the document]
eugener: Currently this does not work as I use actual text to calculate the required text width. 
@maaslalani  Do you know of any API to calculate the screen width of styled text?

borders.go
eugener: Good suggestion... coming out with the next update :)

eugener:
That is a matter of opinion :)  Let's vote on this?

maaslalani:
Yeah, definitely subjective 👍 

Curious what @meowgorithm thinks.

I say this because `lazygit` has this sort of styling:

image

This might be a bit tricky but one thing I could imagine is that `Padding()` on the `BorderTitleStyle` causes the spacing to be done with the `BorderTop` and `Margin()` uses whitespace. But, this might not be worth all the extra complexity, so curious to hear thoughts. However, this would allow us to support both, white spacing and border top padding.

example/main.go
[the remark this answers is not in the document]
meowgorithm: I'd additionally add a rule where, if padding is unset on the style it defaults to `Padding(0, 1)`:

```
╭ Title ───
```

However if a horizontal padding value is set, accept it at face value. For example, `PaddingLeft(0).PaddingRight(3)`

```
╭Title   ───
```

maaslalani:
Hey @eugener, we appreciate the awesome work you've put into this.

We'll play around with this now as it is and some different options we have to achieve this functionality that we were toying around with and review this PR.

One hesitation we have is whether the API is correct, since it does feel a little bit strange having recursively embedded styles within a `lipgloss.Style`. But, maybe this is okay, so we'll play around with this and let you know what we think soo.

eugener:
All the suggestions are implemented. The code is now in your hands (let me know how I can help).
Thank you for all the suggestions again, looking forward to see this functionality in the next release - many users will appreciate this.

eugener:
Is there any plan to merge this or add similar functionality?

bashbunni:
Hey, thanks for your patience on this. We're still deciding what the public API should look like; the current solution in this PR includes some nested styles that we need to consider. In the meantime, we've got this gist that demonstrates how you can achieve this with the existing Lipgloss API.

https://gist.github.com/meowgorithm/1777377a43373f563476a2bcb7d89306

PR #139, opened 2022-10-05
feat: make Lipgloss Style thread-safe

muesli:
Since we had a few reports about concurrently accessing base style definitions, I figured one solution would be making `Style`'s rules map thread-safe. While it's an elegant solution, it is a trade-off: it guards against crashes from concurrent access, but that obviously slows down access by a tiny bit. I don't think the impact is too bad, though, and not crashing is always nice.

```bash
go test -bench=. -benchtime=30s
```

Before change:
```
goos: linux
goarch: amd64
pkg: github.com/charmbracelet/lipgloss
cpu: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
BenchmarkStyleRender-4           7875594              4555 ns/op
PASS
```

With `sync.map`:
```
goos: linux
goarch: amd64
pkg: github.com/charmbracelet/lipgloss
cpu: Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
BenchmarkStyleRender-4           7017492              5167 ns/op
PASS
```

### DIFF
--- a/get.go
+++ b/get.go
@@ -397,12 +397,12 @@ func (s Style) GetFrameSize() (x, y int) {
 
 // Returns whether or not the given property is set.
 func (s Style) isSet(k propKey) bool {
-	_, exists := s.rules[k]
+	_, exists := s.rules.Load(k)
 	return exists
 }
 
 func (s Style) getAsBool(k propKey, defaultVal bool) bool {
-	v, ok := s.rules[k]
+	v, ok := s.rules.Load(k)
 	if !ok {
 		return defaultVal
 	}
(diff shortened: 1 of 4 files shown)


muesli:
Here's one thing I dislike about my own change: this doesn't protect the `value` member of the `Style`. This is typically not a problem, because we operate on a copy of `Style` in critical situations. I can already see future changes invalidating this argument, though, and we would suddenly have to start debugging similar crashes again.

With that being said, maybe an exclusive `sync.Mutex` is the better solution for our problem?

meowgorithm:
We’re deprecating `SetString` in the `instances` branch, right? If so, that would deprecate `value` here, so maybe that’s a non-issue.

Otherwise I think this this is great. It doesn’t appear to impact performance much and the thread-safety it brings is a pretty big deal.

PR #140, opened 2022-10-05
feat: lipgloss renderer

aymanbagabas:
* bump termenv & reflow
* use custom renderers
* update readme
* add examples


color.go
```diff
@@ -1,101 +1,18 @@
...
-	explicitBackgroundColor = true
-}
-
-// TerminalColor is a color intended to be rendered in the terminal. It
-// satisfies the Go color.Color interface.
+// TerminalColor is a color intended to be rendered in the terminal.
 type TerminalColor interface {
-	value() string
-	color() termenv.Color
-	RGBA() (r, g, b, a uint32)
```
muesli, line 96: Not sure if anybody uses this interface, but since it's public we should strive to keep the `RGBA` function in there. You can probably just copy my implementation from the `instances` branch.

renderer.go
```diff
@@ -0,0 +1,144 @@
...
+		output: termenv.DefaultOutput(),
+	}
+	for _, option := range options {
+		option(r)
+	}
+	return r
+}
+
+// WithOutput sets the termenv Output to use for rendering.
+func WithOutput(output *termenv.Output) RendererOption {
```
muesli, line 35: Similar to what we plan in `tea`, we should probably have a `WithOutput` that expects a regular file descriptor and rename this implementation to `WithTermenvOutput`.

muesli:
`examples/layout/go.mod` has an incorrect relative path in the `replace` directive. Maybe we should just keep a single `go.mod` directly in `examples` and not introduce separate definitions per example. Keeping them update is painful enough as it is (see #142 and many other commits).

color.go
aymanbagabas: We'd have to change the signature of `RGBA()` to `RGBA(*Renderer)` though. Without the renderer, we wouldn't be able to detect dark background colors :/ This affects `AdaptiveColor` and `CompleteColor`.

muesli: Correct, but instead of breaking the compatibility, I suggest behaving like before and returning the color value for the default output.

muesli (approved):
Ok, LGTM. We could squash a few commits, but we could also just ignore it and go ahead merge this one now.

PR #143, opened 2022-10-09
fix: Replace  `\n` with a space in Inline style

### LINKED ISSUE #116: Inline concatenates words from multiple lines
When I do `style.Render("hello\nworld")`
I am expecting 2 different words.

If I use `Inline`, the result is `helloworld` without space.

I think we want 2 separate words here, i.e. replace newlines by spaces during inline mode.

Found while working on #109.

griimick:
Fixes #116

This a straightforward modification but I want some clarification,
- Should we trim trailing `\n` as it would add a trailing space?

### DIFF
--- a/style.go
+++ b/style.go
@@ -264,6 +264,7 @@ func (s Style) Render(str string) string {
 
 	// Strip newlines in single line mode
 	if inline {
+		str = strings.Trim(str, "\n")
 		str = strings.ReplaceAll(str, "\n", " ")
 	}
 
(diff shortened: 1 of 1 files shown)


style.go
```diff
@@ -264,6 +264,7 @@ func (s Style) Render(str string) string {
 
 	// Strip newlines in single line mode
 	if inline {
+		str = strings.Trim(str, "\n")
```
muesli, line 267: Just a heads up: this also trims leading newlines. It may or may not be what we want here, I'll still have to think about the implications some more.

PR #147, opened 2022-10-15
docs: minor fixes to comments

winder:
While reviewing the code I came across some typos and stale variable references.

### DIFF
--- a/set.go
+++ b/set.go
@@ -39,7 +39,7 @@ func (s Style) Italic(v bool) Style {
 
 // Underline sets an underline rule. By default, underlines will not be drawn on
 // whitespace like margins and padding. To change this behavior set
-// renderUnderlinesOnSpaces.
+// UnderlineSpaces.
 func (s Style) Underline(v bool) Style {
 	s.set(underlineKey, v)
 	return s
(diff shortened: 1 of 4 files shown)


maaslalani:
Thanks so much!

set.go
```diff
@@ -39,15 +39,15 @@ func (s Style) Italic(v bool) Style {
...
+// UnderlineSpaces.
 func (s Style) Underline(v bool) Style {
 	s.set(underlineKey, v)
 	return s
 }
 
 // Strikethrough sets a strikethrough rule. By default, strikes will not be
 // drawn on whitespace like margins and padding. To change this behavior set
-// renderStrikethroughOnSpaces.
+// StrikethroughSpaces.
```
winder, line 50: These were the only ones which did not seem obvious. After poking around in the history I did not find any reference to `renderUnderlinesOnSpaces` or `renderStrikethroughOnSpaces`, so this is my guess about the correct comment.

maaslalani: I believe your change is correct. Setting the `UnderlineSpaces` and `StrikethroughSpaces` is the right approach, I think the comment was affected by a find + replace. The user would not be able to set the previous methods anyway since they are unexported.

PR #148, opened 2022-10-18
chore: unify get border size function names

nerg4l:
Current functions to access the width of the borders directly on `lipgloss.Style`.

- `GetBorderTopWidth`
- `GetBorderLeftSize`
- `GetBorderBottomSize`
- `GetBorderRightSize`

From this list `GetBorderTopWidth` is a bit of an odd function name compared to the rest because it has different suffix. This PR adds `GetBorderTopSize` and deprecates `GetBorderTopWidth`.


muesli:
Looks reasonable to me!

@meowgorithm Can you confirm this was just an accident and not the intended name?

meowgorithm:
Good catch and, yes, totally an accident. Thanks for the PR @nerg4l!

PR #150, opened 2022-10-19
fix: infinite loop when whitespaceChars width is 0

### LINKED ISSUE #108: Infinite loop when  using tab as whitespace
Found with the following:

```go
Place(10, 3, Center, Center, "hello",
    WithWhitespaceChars("\t"),
  ))
```

lipgloss hangs forever.

cc @meowgorithm

griimick:
Fixes #108

This fixes the infinite loop mentioned in issue.

### DIFF
--- a/whitespace.go
+++ b/whitespace.go
@@ -15,7 +15,7 @@ type whitespace struct {
 
 // Render whitespaces.
 func (w whitespace) render(width int) string {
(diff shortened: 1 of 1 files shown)


griimick:
Unrelated but there are more corner cases to `withWhitespaceChars` like,
- `".\n"`
- `".\t"`

Should we ignore runes with 0 `PrintableRuneWidth` (`\n`, `\t`) when used as `whitespaceChars`?

PR #163, opened 2022-12-11
fix: Align lines with emojis correctly

mikelorant:
Alignment issues occur when some lines have certain emojis in them. This is caused by incorrectly calculating the width of a string.

For a string with emojis in them, it is important to understand that the string width is different to the rune width. Many functions within Lipgloss base the string length on the printable rune width which is incorrect. The correct method requires finding the string width.

The reflow package has a printable rune width function. While the ANSI sequences need to be removed, we do not want the rune width as this function returns.

This change adds a new function to get the printable string width. The code was taken from reflow and modified for use in Lipgloss. Since the printable rune width function in reflow works correctly, it was more reasonable to add the function to Lipgloss.

### DIFF
--- a/align.go
+++ b/align.go
@@ -3,7 +3,6 @@ package lipgloss
 import (
 	"strings"
 
-	"github.com/muesli/reflow/ansi"
 	"github.com/muesli/termenv"
 )
 
@@ -15,7 +14,7 @@ func alignTextHorizontal(str string, pos Position, width int, style *termenv.Sty
 	var b strings.Builder
 
 	for i, l := range lines {
-		lineWidth := ansi.PrintableRuneWidth(l)
+		lineWidth := PrintableStringWidth(l)
 
 		shortAmount := widestLine - lineWidth                // difference from the widest line
 		shortAmount += max(0, width-(shortAmount+lineWidth)) // difference from the total width, if set
--- a/borders.go
+++ b/borders.go
@@ -4,7 +4,6 @@ import (
 	"strings"
 
 	"github.com/mattn/go-runewidth"
-	"github.com/muesli/reflow/ansi"
 	"github.com/muesli/termenv"
 )
 
@@ -384,8 +383,8 @@ func renderHorizontalEdge(left, middle, right string, width int) string {
 		middle = " "
 	}
 
-	leftWidth := ansi.PrintableRuneWidth(left)
-	rightWidth := ansi.PrintableRuneWidth(right)
+	leftWidth := PrintableStringWidth(left)
+	rightWidth := PrintableStringWidth(right)
 
 	runes := []rune(middle)
 	j := 0
@@ -398,7 +397,7 @@ func renderHorizontalEdge(left, middle, right string, width int) string {
 		if j >= len(runes) {
 			j = 0
 		}
-		i += ansi.PrintableRuneWidth(string(runes[j]))
+		i += PrintableStringWidth(string(runes[j]))
 	}
 	out.WriteString(right)
 
--- a/get.go
+++ b/get.go
@@ -3,8 +3,10 @@ package lipgloss
 import (
 	"strings"
 
-	"github.com/muesli/reflow/ansi"
+	"github.com/mattn/go-runewidth"
 )
+
+const marker = '\x1B'
 
 // GetBold returns the style's bold value. If no value is set false is returned.
 func (s Style) GetBold() bool {
@@ -492,11 +494,37 @@ func getLines(s string) (lines []string, widest int) {
 	lines = strings.Split(s, "\n")
 
 	for _, l := range lines {
-		w := ansi.PrintableRuneWidth(l)
+		w := PrintableStringWidth(l)
 		if widest < w {
 			widest = w
 		}
 	}
 
 	return lines, widest
+}
+
+// PrintableStringWidth returns the width of the given string.
+func PrintableStringWidth(s string) int {
+	var ansi bool
+	var sr []rune
+
+	for _, c := range s {
+		if c == marker {
+			// ANSI escape sequence
+			ansi = true
+		} else if ansi {
+			if isTerminator(c) {
+				// ANSI sequence terminated
+				ansi = false
+			}
+		} else {
+			sr = append(sr, c)
+		}
+	}
+
+	return runewidth.StringWidth(string(sr))
+}
+
+func isTerminator(c rune) bool {
+	return (c >= 0x40 && c <= 0x5a) || (c >= 0x61 && c <= 0x7a)
 }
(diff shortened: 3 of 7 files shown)


muesli:
Please do fix this upstream if you found any issues. At first glance I don't see a difference however?

mikelorant:
The key change is the use of `runewidth.StringWidth` instead of `runewidth.RuneWidth`

I will provide a repository shortly showcasing the problem and why this pull request resolves the problem.

mikelorant:
I want to clarify that there is no bug with reflow. It is just providing a different answer to the question. Lipgloss needs to know the printable string width, not the printable rune width. Are you willing to accept a pull request to reflow to add this function? Lipgloss still needs to be updated to use it however.

muesli:
Yeah, definitely willing to accept PRs there, so we don't end up copying similar code all over the place.

muesli:
> I will provide a repository shortly showcasing the problem and why this pull request resolves the problem.

Cool, a few test cases might be nice here.

mikelorant:
Test case with both code and images.

https://github.com/mikelorant/emoji-alignment

Hopefully these results are consistent with other platforms and terminals.

meowgorithm:
Unfortunately, I don't believe this is a valid solution. I'm seeing different results with kitty, however I believe you will also see different results with multi-sequence emojis such as the following. I'm providing both emoji literals and unicode sequences here, which Go will evaluate equally.

```go
// alembic
"⚗️"
"\U00002697\U0000FE0F"
```

```go
// kiss: woman, man, medium-dark skin tone, light skin tone (with zero width joiners)
"👩🏾‍❤️‍💋‍👨🏻"
"\U0001F469\U0001F3FE\U0000200D\U00002764\U0000FE0F\U0000200D\U0001F48B\U0000200D\U0001F468\U0001F3FB"
```

mikelorant:
Tested the two emojis you provided into my example code.

example

I think the complexity was always going to be the terminals and operating system support. However, on a Mac you can see this improves the results.

mikelorant:
Even more interesting is my patch seems to get much closer to handling the 12 byte emoji. Normally the main issue is table borders break. This did not occur which indicates this may just be an issue with calculating maximum line length.
example

If we can identify more emojis that fail this will allow us to build a small bubbletea app that can be used for testing purposes. Its hard to identify all the edge cases.

mikelorant:
Finding some very unusual results when using kitty. Providing some comparison shots with iTerm2 and Terminal for reference.

iTerm2
Terminal
Uploading CleanShot 2022-12-12 at 09.25.15.pkitty

Not sure what all this means except more research is required.

mikelorant:
Kitty has its own `wcwidth` implementation while Terminal and iTerm2 use the system `wcwidth`. This may not be solvable if rendering is done differently by terminals.

meowgorithm:
What I would recommend for now is to simply avoid any emojis that are causing you problems.

The crux of the issue here is that emoji widths must both be correctly calculated in both Go and on the terminal level, and neither side is perfect yet. Even Kitty, which is among the best renderers I've seen, doesn't render every single Emoji correctly. Going further, some desktop environments don't support newer emojis.

We have some tooling in the works that should alleviate the problem a bit on the Go side, but there's not ETA yet.

praagyajoshi:
Got it, thanks for the additional context @meowgorithm.  

> simply avoid any emojis that are causing you problems.

Just a clarification on that front, almost all the emojis I tried are suffering from the same issue.  
Some of them include: ✨, 🌊, 💫, ✅, 🚀

meowgorithm:
@praagyajoshi which terminal, OS, and window manager (if relevant) are you using? As I mentioned this varies widely depending on terminal and can be affected by other OS-level factors.

Also, we still need to test properly but I suspect that switching from `go-runewidth` to `rivo/uniseg` directly under the hood may solve for this, particularly as `rivo/uniseg` now supports both the latest version of Unicode (15.0.0) and the ability to return cell widths.

mikelorant:
I'd like to bring this issue up again based on some feedback from @rivo.

> That's because a single code point (rune) is not always a complete character. That's the whole basis of the uniseg package. The [README](https://github.com/rivo/uniseg) explains this in detail. One example given there is the country flag emoji. These emojis must always consist of two runes. What's the width of only one of these runes then? It's basically undefined because it's an incomplete grapheme cluster. The mattn/go-runewidth package gets this fundamentally wrong and I have tried for a long time to help them do it right but there was never much interest in following up on it.
https://github.com/rivo/uniseg/issues/48#issuecomment-1904871545

Using the rune width anywhere is something we need to be very careful about. In most cases I would expect we should only be using the string width.

It is likely we will not see the full benefits unless we replace `RuneWidth` functions across nearly all of the package as well as its dependencies. This makes this quite a complicated and time consuming task.

My recommendation would be to create a new issue for either `bubbles` or `bubbletea` and determine all the dependencies that would need to be fixed. This would then allow for issues and hopefully pull requests in each of the repositories.

meowgorithm:
Hey! We actually began work on switching to `rivo.StringWidth` today. It's not too daunting of a task as we maintain a fairly tight level of control over our dependency chain, at least as far as rendering is concerned. Here’s a few PRs for it so far:

https://github.com/muesli/reflow/pull/69
https://github.com/muesli/termenv/pull/164
https://github.com/charmbracelet/bubbles/pull/462

mikelorant:
@meowgorithm Been tracking the great work on the `StringWidth` improvements. 👍 

This is why I have now turned my attention to `RuneWidth`. This is just so much more complex because we can't just swap it out. While we can convert the rune to a string we now know it is not always going to be valid.

As you are one of the project maintainers, would be great to get some guidance on where we should start. Something experimental to prove this will give us improvements.

meowgorithm:
So I think ultimately be counting and iterating over graphemes like you suggest. For example, ⚗️ is 6 bytes, 2 runes, and 1 grapheme cluster [playground](https://go.dev/play/p/cXdy8s-7pRh).

I'd also, have a look at [this excellent writeup](https://mitchellh.com/writing/grapheme-clusters-in-terminals) on grapheme clusters in terminals by [Mitchell Hashimoto](https://github.com/mitchellh/). It's the best reference I've seen on this subject.

mikelorant:
I've put up a preliminary pull request for `muesli/reflow` to iterate over grapheme clusters instead of runes. There are so complexities that will likely require further discussion but we can deal with them via an issue on that repository instead.

meowgorithm:
Awesome. For reference that's https://github.com/muesli/reflow/pull/71 (also FYI @muesli is on the Charm team).

@maas, heads up re: `textarea`.

mikelorant:
Closing this pull request as this is going to be fixed with muesli/reflow#71.

PR #174, opened 2023-03-02
Ref: change renderer impl & add Renderer() to style

aymanbagabas:
A renderer determines the terminal color capabilities using Termenv. You can force a specific terminal color profile using `renderer.SetColorProfile()`

Add `style.Renderer()` to return a new style with the given renderer.


meowgorithm:
I wonder if something along the lines of `AttachRenderer()` better describes `Clone()` so that there's no ambiguity between `Clone` and `Copy`.

maaslalani:
👍 to renaming `Clone()` to `AttachRenderer()`, another option would be to have `WithRenderer` (on the style)

```go
func (s Style) WithRenderer(r Renderer) {
  // ...
}
```

meowgorithm:
> another option would be to have `WithRenderer` (on the style)

@aymanbagabas, correct me if I'm wrong, but I believe the use case here is actually to associate styles with a specific renderer, or to change the renderer on a style.

```
style := lipgloss.NewStyle().Reverse()

r := lipgloss.NewRenderer()
style = style.AttachRenderer(r)
```

muesli:
I agree that a different name might be better suited here. My only gripe would be that the imperative `Attach` as the function name's prefix suggests it's changing the renderer on the style, instead of returning a new copy.

meowgorithm:
Just to close the loop here, let’s go with `Renderer` for consistency with the rest of the API.

```go
s := lipgloss.NewStyle()
r := lipgloss.NewRenderer()
s = s.Bold(true).Underline(true).Renderer(r)
```

meowgorithm (approved):

PR #182, opened 2023-03-10
docs: fix typos and clean up comments

gzipChrist:
🤙

### DIFF
--- a/renderer.go
+++ b/renderer.go
@@ -33,7 +33,7 @@ func SetDefaultRenderer(r *Renderer) {
 
 // NewRenderer creates a new Renderer.
 //
-// w will be used to determine the terminal's color capabilities.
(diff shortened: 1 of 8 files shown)


renderer.go
```diff
@@ -33,7 +33,7 @@ func SetDefaultRenderer(r *Renderer) {
 
 // NewRenderer creates a new Renderer.
 //
-// w will be used to determine the terminal's color capabilities.
+// It will be used to determine the terminal's color capabilities.
```
muesli, line 36: `w` is actually referring to the `io.Writer` that gets passed in. As such it's not a typo.

gzipChrist: Ah, my mistake!

muesli:
Thanks again!

PR #197, opened 2023-04-29
fix: Fix bugs with border size calculation

mieubrisse:


meowgorithm (changes requested):
Thanks for this! Do you mind excluding the changes to .gitignore from this PR? We like to keep our PRs very specific, and there are a lot of rules in there (like node stuff) that don’t apply to this type of project.

mieubrisse:
Sure - just pushed. I'll open a separate PR with the VSCode and Idea one (I got hit by Git wanting to version control those)

meowgorithm (approved):

meowgorithm:
Thank you so much for this one!

PR #198, opened 2023-04-30
Expanded Gitignore to prevent accidental committing

mieubrisse:
Per the discussion on #197 , I removed the Node files (but left the exclusion of editor & binary files, so they don't get accidentally checked in and pollute the reflog). LMK what you think @meowgorithm

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,24 @@
 ssh_example_ed25519*
+
+# IntelliJ
+.idea
+*.iws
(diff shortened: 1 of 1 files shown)


meowgorithm:
Thanks for this one, however we like to also keep our .gitignores specific to the project at hand and I'd recommend putting these in your global .gitignore. Sorry if I gave you the wrong impression in the other PR!

PR #210, opened 2023-07-27
fix: renderer race condition

aymanbagabas:
Guard accessing the underlying Termenv output behind a mutex. Multiple goroutines can set/get the color profile causing a race condition.


renderer_test.go
```diff
@@ -33,3 +34,15 @@ func TestRendererWithOutput(t *testing.T) {
 		t.Error("Expected renderer to use true color")
 	}
 }
+
+func TestRace(t *testing.T) {
+	r := NewRenderer(io.Discard)
+
+	for i := 0; i < 100; i++ {
+		t.Run("SetColorProfile", func(t *testing.T) {
+			t.Parallel()
```
caarlos0, line 43: I believe the parent test needs to call parallel too

renderer.go
```diff
@@ -110,6 +120,8 @@ func HasDarkBackground() bool {
 // background. A dark background can either be auto-detected, or set explicitly
 // on the renderer.
 func (r *Renderer) HasDarkBackground() bool {
+	r.mtx.RLock()
```
caarlos0, line 123: should darkbackground have its own mutex?

caarlos0 (approved):

renderer_test.go
aymanbagabas: Not really, `t.Parallel` makes the test run parallel to its _adjacent_ tests. In this case, we want the 100 runs to run in parallel but not the parent with other tests.

caarlos0 (approved):

muesli (approved):
Looks good to me!

PR #212, opened 2023-07-31
fix: cache color profile and background

aymanbagabas:
Reintroduce caching the color profile and background

### DIFF
--- a/renderer.go
+++ b/renderer.go
@@ -16,8 +16,16 @@ var renderer = &Renderer{
 // Renderer is a lipgloss terminal renderer.
 type Renderer struct {
 	output            *termenv.Output
-	hasDarkBackground *bool
-	mtx               sync.RWMutex
+	colorProfile      termenv.Profile
+	hasDarkBackground bool
+
+	getColorProfile      sync.Once
+	explicitColorProfile bool
+
+	getBackgroundColor      sync.Once
+	explicitBackgroundColor bool
+
+	mtx sync.RWMutex
 }
 
 // RendererOption is a function that can be used to configure a [Renderer].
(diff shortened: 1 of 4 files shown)


renderer.go
```diff
@@ -59,7 +67,16 @@ func (r *Renderer) SetOutput(o *termenv.Output) {
 
 // ColorProfile returns the detected termenv color profile.
 func (r *Renderer) ColorProfile() termenv.Profile {
-	return r.output.Profile
+	r.mtx.RLock()
+	defer r.mtx.RUnlock()
+
+	if !r.explicitColorProfile {
+		r.getColorProfile.Do(func() {
+			r.colorProfile = r.output.EnvColorProfile()
```
caarlos0, line 75: should it wlock here?

renderer.go
```diff
@@ -120,10 +139,14 @@ func HasDarkBackground() bool {
 func (r *Renderer) HasDarkBackground() bool {
 	r.mtx.RLock()
 	defer r.mtx.RUnlock()
-	if r.hasDarkBackground != nil {
-		return *r.hasDarkBackground
+
+	if !r.explicitBackgroundColor {
+		r.getBackgroundColor.Do(func() {
+			r.hasDarkBackground = r.output.HasDarkBackground()
```
caarlos0, line 145: should it wlock here?

caarlos0 (approved):
2 questions, but looks good to me

renderer.go
aymanbagabas: It looks like we do, but I think we don't need to since the function would only run once. Maybe @muesli can weigh on this

muesli: Agree, the `sync.Once` already provides its own locking mechanism, so the read-lock should be sufficient. Could maybe add a comment in that regard to prevent further confusion.

muesli (approved):

PR #218, opened 2023-08-26
Feature: Tables

maaslalani:
Adds `NewTable` + features to Lip Gloss.

API:

```go
t := lipgloss.NewTable().
    BorderStyle(lipgloss.NewStyle().Foreground(lipgloss.Color("99"))).
    StyleFunc(func(row, col int) lipgloss.Style {
        switch {
        case row == 0:
            return HeaderStyle
        case row%2 == 0:
            return EvenRowStyle
        default:
            return OddRowStyle
        }
    }).
    Headers("LANGUAGE", "FORMAL", "INFORMAL").
    Row("Chinese", "Nǐn hǎo", "Nǐ hǎo").
    Row("French", "Bonjour", "Salut").
    Row("Russian", "Zdravstvuyte", "Privet").
    Row("Spanish", "Hola", "¿Qué tal?")
```

Results:

```
┌──────────┬──────────────┬───────────┐
│ LANGUAGE │    FORMAL    │ INFORMAL  │
├──────────┼──────────────┼───────────┤
│ Chinese  │ Nǐn hǎo      │ Nǐ hǎo    │
│ French   │ Bonjour      │ Salut     │
│ Russian  │ Zdravstvuyte │ Privet    │
│ Spanish  │ Hola         │ ¿Qué tal? │
└──────────┴──────────────┴───────────┘
```


maaslalani:
image

bashbunni:
I think we should have a "DefaultStyle" function that returns some basic styling for a table

bashbunni:
I think we also need to add some kind of restrictions or a check to make sure that we don't try to index out of range when iterating over the cells in the rows since there is the possibility right now that there are more cells in a row than headings in the header. 
Here is where the issue surfaces if that's the case. Happy to add a test for this case :) 
https://github.com/charmbracelet/lipgloss/pull/218/files#diff-134c48f173480a9772719053c1b3e967756a71c75f84ed4f94f12c71cd91ed30R169

bashbunni:
@maaslalani any thoughts on the above?

bashbunni:
Note to self:
- uses the struct's `String` function when rendering outputs to the table
- use the spread operator to use array values as the `any` type

bashbunni:
This has been great to work with btw Maas!

maaslalani:
Thanks so much @bashbunni!

examples/table/pokemon/main.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+		return data
+	}
+
+	t := table.New().
+		Border(lipgloss.NormalBorder()).
+		BorderStyle(re.NewStyle().Foreground(lipgloss.Color("238"))).
+		Headers(CapitalizeHeaders(headers)...).
+		Rows(data...).
+		StyleFunc(func(row, col int) lipgloss.Style {
```
bashbunni, line 73: @maaslalani I think `StyleFunc` should be named something else to indicate that it's for cell styling. It's a bit vague imo

meowgorithm: Side note that this may become a more general `RenderFunc` (see image).

image-1

bashbunni: Ah gotcha, good to know! Thank you :)

table/rows.go
```diff
@@ -0,0 +1,133 @@
+package table
+
+// Data is the interface that wraps the basic methods of a table model.
+type Data interface {
+	Row(row int) Row
+	Append(row Row)
```
muesli, line 6: I think `Append` should remain optional. Not all data models can/will support that.

table/table.go
```diff
@@ -0,0 +1,564 @@
...
+// style returns the style for a cell based on it's position (row, column).
+func (t *Table) style(row, col int) lipgloss.Style {
+	if t.styleFunc == nil {
+		return lipgloss.NewStyle()
+	}
+	return t.styleFunc(row, col)
+}
+
+// Rows sets the table data.
+func (t *Table) Rows(data Data) *Table {
```
muesli, line 107: I think `Rows(data Data)` is even weirder then `Data(data Data)`. Also comparing `Row` with `Rows` , I would expect them to accept the same type, just singular vs plural?

maaslalani: Yep you're right! I will change this.

I'll keep the Rows method and then also add a Data method.

table/table.go
```diff
@@ -0,0 +1,532 @@
...
+// Rows appends rows to the table data.
+func (t *Table) Rows(rows ...[]string) *Table {
+	if t.data == nil {
+		t.data = NewStringData()
+	}
+	for _, row := range rows {
+		switch t.data.(type) {
+		case *StringData:
+			t.data.(*StringData).Append(StringRow(row))
+		}
```
maaslalani, line 120: @muesli does this seem reasonable to you? This is how we would avoid `Append` being a required message, but not sure if this is the correct way to do it.

PR #223, opened 2023-09-15
Introduce triangle border

nervo:
This looks like this :)

```
◤▼▼▼▼▼▼▼▼◥
▶        ◀
▶  Foo   ◀
▶  Bar   ◀
▶        ◀
◣▲▲▲▲▲▲▲▲◢
```

You could say that's a bit ugly, but such a border unleash its real power when used for styling - for instance - list elements:

```
  Foo
▶ Bar
  Baz
```

In this example, normal elements got a simple `PaddingLeft(2)`, where focus element got `PaddingLeft(1).BorderStyle(lipgloss.TriangleBorder()).BorderLeft(true)`.

### DIFF
--- a/borders.go
+++ b/borders.go
@@ -149,6 +149,17 @@ var (
 		BottomLeft:  " ",
 		BottomRight: " ",
 	}
+
+	triangleBorder = Border{
+		Top:         "▼",
+		Bottom:      "▲",
+		Left:        "▶",
+		Right:       "◀",
+		TopLeft:     "◤",
+		TopRight:    "◥",
+		BottomLeft:  "◣",
+		BottomRight: "◢",
+	}
 )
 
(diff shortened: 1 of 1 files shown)


maaslalani:
Hi @nervo, thanks for the PR and contribution! We haven't seen too much demand for the Triangle border and since we would like to keep the API surface as small as possible I think it is better if this border is added case-by-case into the code as a custom border rather than supporting it in Lip Gloss.

If many more people ask for this border however we would gladly support it and I will re-open this PR!

Thanks again for the awesome contribution!

maaslalani:
For what it's worth, we are planning on adding a List renderer to Lip Gloss which should solve many of your use cases of this border.

https://github.com/charmbracelet/lipgloss/pull/240

PR #224, opened 2023-09-16
Fix GetBorderRightSize

nervo:
It looks like a typo :)


maaslalani:
Thanks so much @nervo!

PR #225, opened 2023-09-16
Fix both GetHorizontalBorderSize and GetVerticalBorderSize

nervo:
In both `GetHorizontalBorderSize` and `GetVerticalBorderSize`, computing size directly using `getBorderStyle()` can not take into accounts that some borders ("top", maybe "right", "bottom", whatever...) could have been disabled, leading to incorrect results.

This is well done in each individual `GetBorder[Top|Right|Bottom|Left]Size`like this: https://github.com/charmbracelet/lipgloss/blob/df8b3fa1f20885458615189fa0558c6226be8240/get.go#L303-L305

So this pull request just make use of this correct behavior :)


maaslalani:
Thanks so much @nervo! We really appreciate the PRs!

PR #228, opened 2023-10-04
Table Data Interface

maaslalani:
- wip: provide model interface for table data
- refactor: rename Model to Data
- refactor: add `Row` method
- refactor: use `Row` method
- fix: examples to Data interface


maaslalani:
Thanks @muesli for all the changes here ❤️

PR #237, opened 2023-10-20
fix: Corrected border shorthand functions explanation

ReidMason:


maaslalani:
Thank you so much! We really appreciate it!

PR #243, opened 2023-12-06
test: last tree node is a subtree

caarlos0:
this is currently resulting in 

```
        ├── Foo
        ├── Bar
        │  ├── Qux
        │  ├── Quux
        │  │  ├── Foo
        │  │  └── Bar
        │  └── Quuux
```

not sure how to best fix it, though... 

I think maybe the api needs be a bit different, e.g., instead of:

```go
New("foo", New("bar", "zaz", New("aa"))
```

something like:

```go
New(New("foo", "bar", New("zaz", "aa")))
```

This is also what I was kinda expecting: the first param is the node name, and the remaining are its children.

It would also make a bit more sense in internal semantics, as currently the children of a node can be either the next nodes or a subtree, which IMHO is the thing that should be called children.

thoughts?

### DIFF
--- a/examples/tree/main.go
+++ b/examples/tree/main.go
@@ -1,0 +1,33 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/charmbracelet/lipgloss/tree"
+)
+
+func main() {
+	fmt.Println(
+		tree.New(
+			"Items",
+			tree.New(
+				"Vegetables",
+				"Lettuce",
+				"Cabbage",
+			),
+			tree.New(
+				"Drinks",
+				"Beer",
+				"Wine",
+				"Whiskey",
+			),
+			"Bread",
+			tree.New(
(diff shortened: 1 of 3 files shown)


caarlos0:
last commit:

- fixes last node of the tree being a subtree
- allows for a tree to have a name and children
- fixes other bugs found 
- changes the rendering so instead of rendering per line, it renders an slice of nodes directly

maaslalani (approved):

maaslalani:
We’ll need to double check that multi-line elements show up correctly, that way we can embed lists within trees, which was the reason for the tree abstraction

caarlos0:
superseeded by https://github.com/charmbracelet/lipgloss/pull/264

caarlos0:
closing in favor of https://github.com/charmbracelet/lipgloss/pull/264

PR #249, opened 2024-01-05
Added option to render styles with optional separator

SuperJura:
The problem with the current implementation is that you cannot modify the rendering of the styles when you have multiple lines to render.
The only way to render multilines is having "\n" between the lines. That is not only cumbersome, but has bugged view with extra space character at the start of every new line.
This PR fixes this.

this is how it looks when rendering code 
`
lines = []string{
		"first line",
		"\n",
		"second line",
	}
`
image

this is how it looks when rendering PR code 
`
lines = []string{
		"first line",
		"second line",
	}
...
style.RenderWithSeparator("\n", lines...),
`
image

### DIFF
--- a/style.go
+++ b/style.go
@@ -110,6 +110,12 @@ func joinString(strs ...string) string {
 	return strings.Join(strs, " ")
 }
 
+// joinString joins a list of strings into a single string separated with a
+// space.
+func joinStringWithSeparator(separator string, strs ...string) string {
+	return strings.Join(strs, separator)
+}
+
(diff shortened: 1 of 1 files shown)


meowgorithm:
Hi! Thanks for this—it's interesting and your implementation is great. We're going to consider it, though do note that we're generally cautious about increasing API surface area.

SuperJura:
great, let me know what you decided.

SuperJura:
any news about this?

Issue #252, opened 2024-01-24
Unable to add right margin to an empty string


mikelorant:
**Describe the bug**
When attempting to add a margin to an empty string, the width is reported as `0`.

**Setup**
 - macOS 12.7.2
 - zsh 5.9
 - iTerm 3.4.23

**To Reproduce**
I have created a repository with documentation, test application and unit tests to demonstrate this issue:

Clone the repository at: https://github.com/mikelorant/lipgloss-width
```
go run main.go
```

To run the tests:
```
go test ./... -v
```

**Source Code**
The most basic way to verify this problem is with this code:
```
str := ""

style := lipgloss.NewStyle().
  Bold(false).
  MarginRight(1)

render := style.Render(str)

log.Println(lipgloss.Width(render))
```

This will output `0`.

Changing `bold` to `true` changes the output to `1`.

Be aware that the terminal capabilities detected by `muesli/termenv` will change the outcome if the terminal cannot provide ANSI capabilities. This can be forced with `export NO_COLOR=1` when enabling bold to see how the outcome changes.

**Expected behavior**
The expected result is `1`.

mikelorant:
@meowgorithm The code for `padRight` was initially added by you. Are you okay with the fix as part of #253?

I'd also like to refactor `padLeft` and `padRight` to share a `pad` function as there is only a 2 line difference between each function.

mikelorant:
I have refactored the padding functions `padLeft` and `padRight` to use a common function called `pad` in #254.

mikelorant:
This issue has been resolved.

PR #253, opened 2024-01-26
Fix adding right padding to empty string

mikelorant:
An empty string could not have padding applied on the right side.

Issue: #252

### DIFF
--- a/style.go
+++ b/style.go
@@ -489,7 +489,7 @@ func padLeft(str string, n int, style *termenv.Style) string {
 
 // Apply right padding.
 func padRight(str string, n int, style *termenv.Style) string {
-	if n == 0 || str == "" {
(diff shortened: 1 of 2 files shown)


mikelorant:
@meowgorithm Can I get approval to run the workflows? Thanks.

mikelorant:
Can I get someone from @charmbracelet to review this. Thanks.

meowgorithm (changes requested):
Nice, this change makes sense to me. Do you mind adding a test as a proof?

mikelorant:
Tests have been added. Needed a little refactor for the test function but now lets us control what text we render. Unit tests all passing.

meowgorithm:
Nice work, LGTM!

PR #254, opened 2024-01-26
Refactor padding functions

mikelorant:
The `padLeft` and `padRight` functions were nearly identical and could benefit from being refactored to share common code.

A new function called `pad` was added that applies padding based on the sign of the amount of padding needed. A positive value applies padding to the right side while a negative value applies padding to the left side.

Issue: #252

### DIFF
--- a/style.go
+++ b/style.go
@@ -464,36 +464,23 @@ func (s Style) applyMargins(str string, inline bool) string {
 
 // Apply left padding.
 func padLeft(str string, n int, style *termenv.Style) string {
-	if n == 0 {
-		return str
-	}
-
-	sp := strings.Repeat(" ", n)
-	if style != nil {
-		sp = style.Styled(sp)
-	}
-
-	b := strings.Builder{}
-	l := strings.Split(str, "\n")
-
-	for i := range l {
-		b.WriteString(sp)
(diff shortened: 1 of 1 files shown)


mikelorant:
@meowgorithm Can I get approval to run the workflows? Thanks.

meowgorithm:
Done. Nice work on this one.

mikelorant:
Can I get someone from @charmbracelet to review this. Thanks.

meowgorithm:
Good call on consolidating the two functions. If I'm reading this correctly negative integers applied to `pad` pad the left side, while positive integers pad the right, correct?

mikelorant:
> If I'm reading this correctly negative integers applied to pad pad the left side, while positive integers pad the right, correct?

Yes, you reading this correctly.

I am doing the conflict resolution now and will have this updated shortly.

mikelorant:
Added function comment to explain the relationship between positive and negative values.

mikelorant:
@meowgorithm Any improvements or changes you'd like to make?

meowgorithm:
Looking good, @mikelorant—let's get it done. Thanks for the attention to detail!

PR #255, opened 2024-01-29
Apply gofumpt to all files

mikelorant:
Apply gofumpt to all files to correct minor formatting issues.

### DIFF
--- a/align.go
+++ b/align.go
@@ -69,7 +69,7 @@ func alignTextVertical(str string, pos Position, height int, _ *termenv.Style) s
 	case Top:
 		return str + strings.Repeat("\n", height-strHeight)
 	case Center:
-		var topPadding, bottomPadding = (height - strHeight) / 2, (height - strHeight) / 2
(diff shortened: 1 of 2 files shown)


mikelorant:
Can I get someone from @charmbracelet to run this workflow and review this. Simple fix that doesn't need much explanation. Thanks.

align.go
```diff
@@ -69,7 +69,7 @@ func alignTextVertical(str string, pos Position, height int, _ *termenv.Style) s
 	case Top:
 		return str + strings.Repeat("\n", height-strHeight)
 	case Center:
-		var topPadding, bottomPadding = (height - strHeight) / 2, (height - strHeight) / 2
+		topPadding, bottomPadding := (height-strHeight)/2, (height-strHeight)/2
```
meowgorithm, line 72: Looks good, @mikelorant, however can you add a nolint directive for the magic number `2` here?

meowgorithm (changes requested):

meowgorithm:
Awesome, thank you!

PR #263, opened 2024-02-15
feat: use tree impl for lists

caarlos0:
This will use the tree renderer to render lists too.
Both are very similar if you think about it, the only difference is how they are displayed.

I ended up moving a lot of stuff around, and removing some other things - mainly filter, hide, etc, which I think are more behavior-related than styling, so maybe they should be in the bubble that uses list instead of here?

Anyway, let me know what you think!

PS: all tests are creating golden files, which you can inspect to see how the output looks like :)


caarlos0:
I call this "the frankenlistre"

CleanShot 2024-02-15 at 16 19 40@2x

caarlos0:
todo

- [ ] looking into the examples, probably will have to implement `At(int)`

caarlos0:
superseeded by https://github.com/charmbracelet/lipgloss/pull/264

caarlos0:
closing in favor of https://github.com/charmbracelet/lipgloss/pull/264

PR #242, opened 2023-11-30
Tree Renderer

maaslalani:
Implements `tree` renderer.

```go
tree := New(
  "Foo",
  "Bar",
  New(
    "Qux",
    "Quux",
    New(
      "Foo",
      "Bar",
    ),
    "Quuux",
  ),
  "Baz",
)
```

Produces:

```
├── Foo
├── Bar
│  ├── Qux
│  ├── Quux
│  │  ├── Foo
│  │  └── Bar
│  └── Quuux
└── Baz
```

PR #240, opened 2023-11-21
List Renderer

maaslalani:
## Rendering Lists

Lip Gloss ships with a list rendering sub-package.

```go
import "github.com/charmbracelet/lipgloss/list"
```

Define a new list.

```go
l := list.New("A", "B", "C")
```

Print the list.

```go
fmt.Println(l)

// • A
// • B
// • C
```

### Customization

Lists can be customized via their enumeration function as well as using
`lipgloss.Style`s.

```go
enumeratorStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("99")).MarginRight(1)
itemStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("212")).MarginRight(1)

l := list.New(
  "Glossier",
  "Claire’s Boutique",
  "Nyx",
  "Mac",
  "Milk",
).
  Enumerator(list.Roman).
  EnumeratorStyle(enumeratorStyle).
  ItemStyle(itemStyle)
```

Print the list.

<p align="center">
List example
</p>

In addition to the predefined enumerators (`Arabic`, `Alphabet`, `Roman`, `Bullet`, `Tree`),
you may also define your own custom enumerator:

```go
var DuckDuckGooseEnumerator Enumerator = func(l *List, i int) string {
    if l.At(i) == "Goose" {
        return "Honk →"
    }
    return ""
}
```

Use it in a list:

```go
l := list.New("Duck", "Duck", "Duck", "Duck", "Goose", "Duck", "Duck")
l.Enumerator(DuckDuckGooseEnumerator)
```

Print the list:

<p align="center">
image
</p>

### Building

If you need, you can also build lists incrementally:

```go
l := list.New()

for i := 0; i < repeat; i++ {
    l.Item("Lip Gloss")
}
```

PR #264, opened 2024-02-16
feat: trees and lists

caarlos0:
Worked on top of @maaslalani trees and lists implementations, fixed some bugs, refactor some stuff, added a couple of possibilities (namely, sublists (as trees)), etc.

Its a big PR, but it if it makes it better, it closes this 4 other PRs:

closes #242 
closes #240 
closes #243 
closes #263 

Some examples:

CleanShot 2024-02-16 at 10 30 22@2x
CleanShot 2024-02-16 at 10 31 00@2x
CleanShot 2024-02-16 at 10 31 25@2x
CleanShot 2024-02-16 at 10 32 43@2x
CleanShot 2024-02-16 at 11 08 51@2x


maaslalani:
This is _sooooo_ awesome @caarlos0. Thank you so much!

We'll need to refactor this to make the API much simpler again however. We should take out the concept of "Sublists" and "Renderers".

Everything should be a List or Tree (even sublists and subtrees):


So the following example (in the test):

```go
list.New().
	Item("Foo").
	Item(list.NewSublist("Bar", "foo2", "bar2")).
	Item(
		list.NewSublist("Qux", "aaa", "bbb").
			Renderer(
				list.NewDefaultRenderer().
					EnumeratorStyle(style1).
					Enumerator(list.Roman),
			),
	).
	Item(
		list.NewSublist("Deep").
			Renderer(
			list.NewDefaultRenderer().
				EnumeratorStyle(style2).
				Enumerator(list.Alphabet),
		).
		Item("foo").
		// ...

```

Should have an API of the following:

```go
list.New().
	Item("Foo").
	Item(list.New("Bar", "foo2", "bar2")).
	Item(list.New("Qux", "aaa", "bbb").EnumeratorStyle(style1).Enumerator(list.Roman)).
	Item(list.New("Deep").EnumeratorStyle(style2).Enumerator(list.Alphabet)).
	Item("foo").
	// ...
```

caarlos0:
thanks

caarlos0:
TODO:

- [ ] update readme (verify list, add trees)

examples/list/duckduckgoose/main.go
```diff
@@ -0,0 +1,27 @@
...
+
+import (
+	"fmt"
+
+	"github.com/charmbracelet/lipgloss"
+	"github.com/charmbracelet/lipgloss/list"
+	"github.com/charmbracelet/lipgloss/tree"
+)
+
+func duckDuckGooseEnumerator(data tree.Data, i int, _ bool) (string, string) {
```
maaslalani, line 11: I'm not a big fan of how a `tree.Data` is passed here for the enumerator. I believe this should simply be passed the `list` so that users don't confuse trees and lists.

maaslalani: I think the enumerator API should simply be:

```go
func (l list.List, i int) string
```

Perhaps even (since you could theoretically query the list separately given the index, but not a huge fan of this personally):

```go
func (i int) string
```

Is there a possible way to achieve that?

maaslalani: For the `last bool`, I believe this information could live on the `tree` and query-able by the index:

Something along the lines of (with a better name):
```
tree.IsLastChild(i)
```

examples/tree/simple/main.go
```diff
@@ -0,0 +1,20 @@
...
+func main() {
+	t := tree.New(
+		"Root",
+		"Item 1",
+		"Item 2",
+		tree.New(
+			"Item 3",
+			"Item 3.1",
+		),
+	)
```
maaslalani, line 18:
```suggestion
	t := tree.New("Root").Item("Item 1", "Item 2", tree.New("Item 3").Item("3.1"))
```

maaslalani: Perhaps `Item` → `Child` for trees.

caarlos0: its the same api for trees and lists though... unless we embed a tree in a `type List struct` and delegate the methods to it 🤔

examples/tree/readme-custom-enumerator/main.go
```diff
@@ -0,0 +1,25 @@
...
+		"",
+		"Glossier",
+		"Claire’s Boutique",
+		tree.New(
+			"Nyx",
+			"Qux",
+			"Quux",
+		),
+		"Mac",
+		"Milk",
```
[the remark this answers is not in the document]
caarlos0: can do yeah, but notice the first arg would have to be mandatory... or we make something like

```go
tree.New().Root("foo").Item("bar")
```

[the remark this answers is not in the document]
maaslalani: Yes or variadic arguments for the root 🤔

examples/list/duckduckgoose/main.go
caarlos0: The `last bool` was indeed not needed anymore, removed it.

As for passing the list directly, I disagree for two reasons:

1. there is no `list.List` type, just a `list.New` that returns `tree.Tree`
2. the `tree.Tree` api is too large, IMO we should pass the smallest interface possible in order to do the job, which I think is is `Data` - or something even smaller

examples/tree/readme-custom-enumerator/main.go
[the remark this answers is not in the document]
caarlos0: I think variadic args for root might lead to confusion i.e. what to do when more than one arg was given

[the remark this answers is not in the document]
caarlos0: done

examples/tree/simple/main.go
caarlos0: done

testinfected:
Hey Charm team, this is awesome. When can we expect it to land into master? That's exactly what we're missing.

Thanks!

tree/item.go
```diff
@@ -0,0 +1,110 @@
...
+	m.data = m.data.Append(item)
+	return m
+}
+
+// NewFilter initializes a new Filter.
+func NewFilter(data Data) *Filter {
+	return &Filter{data: data}
+}
+
+// Remove implements Data.
```
bashbunni, line 73:
```suggestion
// Remove allows Filter to implement Data. It removes an element from the Tree.
```

tree/item.go
```diff
@@ -0,0 +1,110 @@
...
+
+var _ Data = NewFilter(nil)
+
+// Filter applies a filter on some data.
+type Filter struct {
+	data   Data
+	filter func(index int) bool
+}
+
+// Append implements Data.
```
bashbunni, line 62:
```suggestion
// Append allows Filter to implement Data. It adds an element to the Tree.
```

tree/item.go
```diff
@@ -0,0 +1,110 @@
...
+				return m.data.At(i)
+			}
+			j++
+		}
+	}
+
+	return nil
+}
+
+// Filter applies the given filter function to the data.
```
bashbunni, line 95:
```suggestion
// Filter uses a filter function to set a condition that all of the data must satisfy to be in the Tree.

File: list/enumerations.go [hunk 7]
```diff
@@ -0,0 +1,86 @@
...
+func Asterisk(Data, int) string {
+	return "*"
+}
+
+// Dash is an enumeration using dashes.
+//
+// * Foo
+// * Bar
+// * Baz
+// * Qux.
```
aymanbagabas, line 83:
```suggestion
// - Foo
// - Bar
// - Baz
// - Qux.
```

aymanbagabas (approved):
Lgtm

tree/item.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, d := range data {
+		s := StringNode(d)
+		result = append(result, &s)
+	}
+	return nodeData(result)
+}
+
+var _ Data = NewFilter(nil)
+
+// Filter applies a filter on some data.
```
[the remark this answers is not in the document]
caarlos0:
```suggestion
// Filter applies a filter on some data. You could use this to create a new 
// tree whose values all satisfy the condition provided in the Filter() function.
```

bashbunni:
Hey @testinfected we're still doing some usability tests on the API to decide if the developer experience is where we want it to be before release. No ETA yet, but the functionality is there, just want to make sure it's easy to use before publishing. We don't want to have to break the API in the future, so now's the time to get it all sorted :)

nervo:
@bashbunni i just wonder something 🤔

There are today two ways to use a list as a user input in the charmbracelet ecosystem:
- the list bubble https://github.com/charmbracelet/bubbles/tree/master/list
- the select/multiselect huh https://github.com/charmbracelet/huh#select
None of them seems to have a compatible interface with this new list lipgloss component :(

Do you plan something on this subject ? It could be amazing to use the same list either in a user input or output way ❤️

caarlos0:
hey @nervo !

Ideally, we'll use this new component on both of these once this is merged :)

bashbunni:
Thoughts:
- We should probably populate the "Overview" section of the Godoc with a basic example
- It's still unclear to me when to use a tree vs a list - maybe something we can include in the godoc?
- If I add sublists in `Items` instead of as an `Item`, I get multiple bullet points.
double-bullets

bashbunni:
We should also improve godoc on `StyleFunc` usage for lists/trees. I'm searching examples to figure out how to use it

caarlos0:
> If I add sublists in Items instead of as an Item, I get multiple bullet points.

can you share an example?

bashbunni:
@caarlos0 🤦‍♀️ totally didn't include the code to repro did I? LOL
```go
	l := list.New().Items(
		"S",
		list.New().Items([]string{"vim", "doom emacs"}),
		list.New().Item("I like fuzzy socks"),
	)

	fmt.Println(l)
```
gives me 
image

caarlos0:
fixed @bashbunni

PR #269, opened 2024-03-15
feat!: introduce color profiles and use term/ansi for styles

aymanbagabas:
This introduces color profiles similar to Termenv color profiles. Plus, it switches to using Charmbracelet term & term/ansi to query terminal background color for the default renderer, and construct styles.

However, it breaks the renderer API since it doesn't depend on Termenv anymore.

```go
func NewRenderer(cp Profile, hasDarkBackground bool) *Renderer

// New exported types
// Profile is a color profile: NoTTY, Ascii, ANSI, ANSI256, or TrueColor.
type Profile int

const (
	// NoTTY, not a terminal profile
	NoTTY Profile = iota
	// Ascii, uncolored profile
	Ascii // nolint: revive
	// ANSI, 4-bit color profile
	ANSI
	// ANSI256, 8-bit color profile
	ANSI256
	// TrueColor, 24-bit color profile
	TrueColor
)

// New exported funcs
func DetectColorProfile(stdout *os.File, environ []string) Profile
func EnvColorProfile(environ []string) Profile
func EnableLegacyWindowsANSI(out *os.File)
```

This also introduces a new color profile `NoTTY`. This profile doesn't generate any ANSI sequences, and, as the name suggests, is used when the renderer is dealing with a non-terminal output. The `Ascii` profile on the other hand is used to generate all the style sequences except color ones. This works well when `NO_COLOR`.


color_test.go
```diff
@@ -3,38 +3,36 @@ package lipgloss
...
-			termenv.Ascii,
+			Ascii,
 			"hello",
 		},
 		{
 			"ansi",
-			termenv.ANSI,
-			"\x1b[94mhello\x1b[0m",
+			ANSI,
+			"\x1b[94mhello\x1b[m",
```
caarlos0, line 25: should these still end with `[0m`?

examples/go.mod
```diff
@@ -4,9 +4,12 @@ go 1.17
 
 replace github.com/charmbracelet/lipgloss => ../
 
+replace github.com/charmbracelet/x/exp/term => ../../x/exp/term
```
caarlos0, line 7: rm before merging

go.mod
```diff
@@ -4,16 +4,17 @@ retract v0.7.0 // v0.7.0 introduces a bug that causes some apps to freeze.
 
 go 1.18
 
+replace github.com/charmbracelet/x/exp/term => ../x/exp/term
```
caarlos0, line 7: same here

profile.go
```diff
@@ -0,0 +1,415 @@
...
+
+// Color creates a Color from a string. Valid inputs are hex colors, as well as
+// ANSI Color codes (0-15, 16-255).
+func (p Profile) Color(s string) ansi.Color {
+	if len(s) == 0 {
+		return noColor
+	}
+
+	var c ansi.Color
+	if strings.HasPrefix(s, "#") {
```
caarlos0, line 71: I think go-colorful (which is already a dep) can do this as well? maybe remove go-colorful, or use it here too...

renderer.go
```diff
@@ -1,38 +1,53 @@
...
-// RendererOption is a function that can be used to configure a [Renderer].
-type RendererOption func(r *Renderer)
-
 // DefaultRenderer returns the default renderer.
 func DefaultRenderer() *Renderer {
+	rendererOnce.Do(func() {
+		if renderer != nil {
+			// Alredy set by SetDefaultRenderer
+			return
+		}
```
caarlos0, line 31: maybe instead do `rendererOnce.Do(func() { renderer = r })` on `SetRenderer`, then this would not be needed anymore 🤔

style_test.go
```diff
@@ -61,44 +58,42 @@ func TestStyleRender(t *testing.T) {
...
 		},
 		{
 			r.NewStyle().Faint(true),
-			"\x1b[2mhello\x1b[0m",
+			"\x1b[2mhello\x1b[m",
 		},
 		{
 			NewStyle().Faint(true).Renderer(r),
-			"\x1b[2mhello\x1b[0m",
+			"\x1b[2mhello\x1b[m",
```
caarlos0, line 96: shouldn't all these strings end with `[0m`?

runes_test.go
```diff
@@ -19,31 +19,31 @@ func TestStyleRunes(t *testing.T) {
...
 			[]int{1, 3},
-			"h\x1b[7me\x1b[0ml\x1b[7ml\x1b[0mo",
+			"h\x1b[7me\x1b[ml\x1b[7ml\x1b[mo",
 		},
 		{
 			"你好 0,1",
 			"你好",
 			[]int{0, 1},
-			"\x1b[7m你好\x1b[0m",
+			"\x1b[7m你好\x1b[m",
```
caarlos0, line 46: here too, `[0m`?

caarlos0 (approved):

color_test.go
aymanbagabas: Short answer: they're the same, `[m` is shorter and consumes one less byte
Long answer: SGR style sequences default value is `0` meaning terminals treat a missing parameter for SGR i.e. `CSI p...p m` sequences as zero. Thus, `CSI m` is equal to `CSI 0 m`

profile.go
aymanbagabas: Good point

color_test.go
caarlos0: oh, TIL!

thanks for explaining!

bashbunni:
Nice! The latest changes fixed the Bubble Tea `chat` example for me. It's now using the right color profile instead of ascii. Tested all of the bubble tea examples last week and they all worked great even before the fix.

meowgorithm:
Closing this as these changes have now happened via other PRs. Cheers!

PR #273, opened 2024-03-18
chore: stub out a few wordwrap tests

meowgorithm:


muesli:
I think this test belongs to the library that implements the functionality.

meowgorithm:
Ah, you're probably right, though I do wonder if we want both, since Lip Gloss adds trailing spaces to build blocks.

```go
str := lipgloss.NewStyle().Width(10).Render("Hello, Kätzchen")
// Hello,••••
// Kätzchen••
```

muesli:
> Ah, you're probably right, though I do wonder if we want both, since Lip Gloss adds trailing spaces to build blocks.

That's a good point, even though we're technically not testing the wrapping then, but padding the individual lines. Maybe just a slight misnomer then, even tho wrapping is involved.

meowgorithm:
Okay, great point. I'll close the PR.

PR #278, opened 2024-04-06
fix: Place strings properly on non-edges

blazingbbq:
Fixes `Place`ing strings on non-edges (Top/Bottom, Left/Right). 
`totalGap - split` is actually the far side value (`Right` for horizontal, `Bottom` for vertical).

✅ Added tests to confirm behavior.

### DIFF
--- a/place_test.go
+++ b/place_test.go
@@ -1,0 +1,41 @@
+package lipgloss
+
+import "testing"
+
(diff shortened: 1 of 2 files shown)


meowgorithm:
Before we get into this I just want to say that I absolutely love your handle.

blazingbbq:
Bump on this

PR #279, opened 2024-04-13
chore(deps): Upgrade dependencies

mikelorant:
Upgrade dependencies to latest stable versions.

As part of this dependency upgrade, the `term` package is being updated to include an important fix for wrapping strings containing ANSI codes. See charmbracelet/x/issues/58 for more details.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,7 @@ retract v0.7.0 // v0.7.0 introduces a bug that causes some apps to freeze.
 go 1.18
 
 require (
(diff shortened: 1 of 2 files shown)


mikelorant:
@aymanbagabas @muesli I have confirmed with my own test app that the ANSI wrapping issue has been fixed so we can upgrade `term` package to the latest code.

I'd also recommend that this would be a good point to create a new Lip Gloss release.

aymanbagabas (approved):

mikelorant:
@aymanbagabas Can we also get `term` replacing `reflow` in Bubbles? I think have a mixture would be a bad idea.

Issue #281, opened 2024-04-22
`Style.GetFrameSize()` does not account for border when using only `BorderStyle()`


qualidafial:
**Describe the bug**

When rendering, a style configured with `BorderStyle(border)` renders the same as using `Border(border)` with no `sides` arguments, thanks to this check in `applyBorder()`:

```go
if border != noBorder && !(topSet || rightSet || bottomSet || leftSet) {
	hasTop = true
	hasRight = true
	hasBottom = true
	hasLeft = true
}
```

However, the border sizes reported from `GetHorizontalBorderSize()`, `GetVerticalBorderSize()`, `GetHorizontalFrameSize()`, and `GetVerticalFrameSize()` are not adjusted when using `BorderStyle()`. I'm new to the Bubbletea ecosystem and have been trying to use these functions to calculate content sizes, and kept getting frames that exceeded the intended size without understanding why.

**To Reproduce**

```go
package lipgloss

import "testing"

func TestStyle_GetBorderSizes(t *testing.T) {
	tests := []struct {
		name  string
		style Style
		wantX int
		wantY int
	}{
		{
			name:  "Default style",
			style: NewStyle(),
			wantX: 0,
			wantY: 0,
		},
		{
			name:  "Border(NormalBorder())",
			style: NewStyle().Border(NormalBorder()),
			wantX: 2,
			wantY: 2,
		},
		{
			name:  "Border(NormalBorder(), true)",
			style: NewStyle().Border(NormalBorder(), true),
			wantX: 2,
			wantY: 2,
		},
		{
			name:  "Border(NormalBorder(), true, false)",
			style: NewStyle().Border(NormalBorder(), true, false),
			wantX: 0,
			wantY: 2,
		},
		{
			name:  "Border(NormalBorder(), true, true, false)",
			style: NewStyle().Border(NormalBorder(), true, true, false),
			wantX: 2,
			wantY: 1,
		},
		{
			name:  "Border(NormalBorder(), true, true, false, false)",
			style: NewStyle().Border(NormalBorder(), true, true, false, false),
			wantX: 1,
			wantY: 1,
		},
		{
			name:  "BorderTop(true).BorderStyle(NormalBorder())",
			style: NewStyle().BorderTop(true).BorderStyle(NormalBorder()),
			wantX: 0,
			wantY: 1,
		},
		{
			name:  "BorderStyle(NormalBorder())",
			style: NewStyle().BorderStyle(NormalBorder()),
			wantX: 2,
[...]

PR #282, opened 2024-04-22
fix: return consistent border sizes when BorderStyle is used.

qualidafial:
Return non-zero border sizes when border style is set, and no border sides have been specifically turned on or off.

Fixes: https://github.com/charmbracelet/lipgloss/issues/281

### DIFF
--- a/borders.go
+++ b/borders.go
@@ -228,11 +228,6 @@ func HiddenBorder() Border {
 
 func (s Style) applyBorder(str string) string {
 	var (
-		topSet    = s.isSet(borderTopKey)
-		rightSet  = s.isSet(borderRightKey)
-		bottomSet = s.isSet(borderBottomKey)
-		leftSet   = s.isSet(borderLeftKey)
-
 		border    = s.getBorderStyle()
 		hasTop    = s.getAsBool(borderTopKey, false)
 		hasRight  = s.getAsBool(borderRightKey, false)
@@ -252,7 +247,7 @@ func (s Style) applyBorder(str string) string {
 
 	// If a border is set and no sides have been specifically turned on or off
 	// render borders on all sides.
-	if border != noBorder && !(topSet || rightSet || bottomSet || leftSet) {
+	if s.autoEnableBorder() {
 		hasTop = true
 		hasRight = true
 		hasBottom = true
@@ -372,6 +367,19 @@ func (s Style) applyBorder(str string) string {
 	}
 
 	return out.String()
+}
+
+func (s Style) autoEnableBorder() bool {
+	// If a border style is set and no sides have been specifically turned on
+	// or off, render borders on all sides.
+	var (
+		border    = s.getBorderStyle()
+		topSet    = s.isSet(borderTopKey)
+		leftSet   = s.isSet(borderLeftKey)
+		rightSet  = s.isSet(borderRightKey)
+		bottomSet = s.isSet(borderBottomKey)
+	)
+	return border != noBorder && !(topSet || rightSet || bottomSet || leftSet)
 }
 
 // Render the horizontal (top or bottom) portion of a border.
--- a/get.go
+++ b/get.go
@@ -300,7 +300,7 @@ func (s Style) GetBorderTopWidth() int {
 // runes of varying widths, the widest rune is returned. If no border exists on
 // the top edge, 0 is returned.
 func (s Style) GetBorderTopSize() int {
-	if !s.getAsBool(borderTopKey, false) {
+	if !s.getAsBool(borderTopKey, s.autoEnableBorder()) {
 		return 0
 	}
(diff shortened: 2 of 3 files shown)


borders.go
```diff
@@ -374,6 +369,19 @@ func (s Style) applyBorder(str string) string {
 	return out.String()
 }
 
+func (s Style) autoEnableBorder() bool {
```
meowgorithm, line 372: Might call this something like `shouldAutoEnableBorder` just to make it obvious it's a check and not changing state, i.e. actually enabling borders

qualidafial: will do

qualidafial: Pushed.

qualidafial:
@meowgorithm ping. I've pushed the changes you requested. Would you mind taking another look?

qualidafial:
Ping?

get.go
```diff
@@ -300,7 +300,7 @@ func (s Style) GetBorderTopWidth() int {
 // runes of varying widths, the widest rune is returned. If no border exists on
 // the top edge, 0 is returned.
 func (s Style) GetBorderTopSize() int {
-	if !s.getAsBool(borderTopKey, false) {
+	if !s.getAsBool(borderTopKey, s.shouldAutoEnableBorder()) {
```
bashbunni, line 303: I think this should probably either short-circuit with `s.shouldAutoEnableBorder` or continue to have `false` as the second arg here. I don't think this function call is necessary unless I'm missing something :thinking: 

e.g. 
```go
if !s.shouldAutoEnableBorder() || !s.getAsBool(borderTopKey, false)
```

Although, again not sure this check saves us in efficiency

bashbunni: If it's to account for the border value set in `BorderStyle`, wouldn't that be handled already at line 306?

bashbunni:
Also related: (and might eliminate the need for multiple checks of `BorderStyle`)
https://github.com/charmbracelet/lipgloss/pull/380/files
https://github.com/charmbracelet/lipgloss/issues/353

I tried adding a custom border in the test introduced here and it's not fixed by this PR (doesn't work on `master` either). I do still think this is a step in the right direction though

Here's the failing test case in case you're curious:

```go
 		{
			name:  "Custom BorderStyle",
			style: NewStyle().BorderStyle(Border{Left: "      new"}),
			wantX: Width("      new"),
			wantY: 0,
		},

```

get.go
qualidafial: The double negative `!a || !b` style is confusing (and I think it should actually be `&&` not `||`). However I think if we invert the `if` condition it will be easier to follow:

```go
if s.shouldAutoEnableBorder() || s.getAsBool(borderTopKey, false) {
	return s.getBorderStyle().GetTopSize()
}
return 0
```

It breaks the Go idiom of putting the happy path return at the bottom, but I think the improvement in clarity is worth it.

qualidafial: > If it's to account for the border value set in `BorderStyle`, wouldn't that be handled already at line 306?

That would ignore the show/hide border flags set on the style, e.g. under `borderTopKey`, which would be a regression I believe. At the very least it would change the behavior of many existing programs which users could rightfully consider a regression.

qualidafial:
> Also related: (and might eliminate the need for multiple checks of `BorderStyle`) https://github.com/charmbracelet/lipgloss/pull/380/files #353
> 
> I tried adding a custom border in the test introduced here and it's not fixed by this PR (doesn't work on `master` either). I do still think this is a step in the right direction though
> 
> Here's the failing test case in case you're curious:
> 
> ```go
>  		{
> 			name:  "Custom BorderStyle",
> 			style: NewStyle().BorderStyle(Border{Left: "      new"}),
> 			wantX: Width("      new"),
> 			wantY: 0,
> 		},
> ```

Based on my read of `Style.applyBorder`, it looks like successive runes in a left or right border are meant to be rendered vertically. e.g. rendering the border `Border{Left:"|:",Right:"<>"}` around the string "foo\nbar\nbaz" would look like:

```
|foo<
:bar>
|baz<
```

So I think `wantX: 1,` is the appropriate expectation from the test case above.

qualidafial:
@bashbunni @meowgorithm I've pushed additional changes based on your comments.

meowgorithm:
Thanks for your patience and effort with this one, @qualidafial. It occurred to me (and @bashbunni): is this basically the same issue as https://github.com/charmbracelet/lipgloss/pull/411? If so we can probably go with this one, but bring in the getters on my branch.

bashbunni:
Thank you so much for your fixes here. I merged the work here with https://github.com/charmbracelet/lipgloss/pull/411 which is now merged into `master` :)

qualidafial:
Thank you for fixing this! :heart:

meowgorithm:
@qualidafial thank you for your patience with this one!

PR #285, opened 2024-04-29
feat: support inferring the color profile from terminfo dbs

aymanbagabas:
This will try to load the terminal terminfo database and infer the color profile supported by the terminal.

### DIFF
--- a/env.go
+++ b/env.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/charmbracelet/x/exp/term"
+	"github.com/xo/terminfo"
 )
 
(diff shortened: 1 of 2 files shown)


bashbunni:
Correct me if I'm wrong, but I think this is our old design idea? in v2 we're not doing any colour detection, though maybe this is an enhancement for v1?

aymanbagabas:
> Correct me if I'm wrong, but I think this is our old design idea? in v2 we're not doing any colour detection, though maybe this is an enhancement for v1?

Right, we don't need this anymore. Closing

PR #291, opened 2024-05-09
fix: Change the propkeys from int to int64

hugoleodev:
When I create binaries for each OS using goreleaser, the propKeys
generates an int overflow since the iota duplicates for each propKey

This commit changes the int type to works well with binaries generated for arm and i386 archs.


aymanbagabas (approved):
Nice catch @hugoleodev!

PR #296, opened 2024-05-15
docs: improve godoc for trees and lists

bashbunni:
WIP: improving godoc given my experience checkin' out the API

- [x] provide basic usage examples in godoc
- [x] explain parameters if less obvious (`StyleFunc`)

--- 

Viewing changes:
`godoc -http=localhost:6060` then connect to `localhost:6060` in browser. From there, look at the list and tree godoc

### DIFF
--- a/tree/tree.go
+++ b/tree/tree.go
@@ -117,10 +117,10 @@ func (n *Tree) String() string {
 //
 // The resulting tree would be:
 //
-// ├── foo
-// ├── bar
-// │   └── zaz
-// └── qux
+//	├── foo
+//	├── bar
+//	│   └── zaz
+//	└── qux
 //
 // You may also change the tree style using Enumerator.
 func (n *Tree) Item(item any) *Tree {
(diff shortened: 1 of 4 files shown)


tree/tree.go
```diff
@@ -1,3 +1,27 @@
...
+//		Item(
+//		tree.New().
+//			Root("Item 3").
+//			Item("Item 3.1").
+//			Item("Item 3.2"),
+//		)
+//
+//	fmt.Println(t)
+//
+// If you're looking to create a list, you can use the list package which wraps
```
bashbunni, line 22: could probably add more info here on the difference between the two, but I *think* this one sentence + the examples showing some differences in implementation might be enough?

bashbunni:
I think the `StyleFunc` examples explain the parameters well enough... If it's unclear going forward, then can always add to the docs later

aymanbagabas (approved):

PR #302, opened 2024-05-24
fix: simplify color profile dectection

caarlos0:
prevents going through the env color detection a second time if clicolor_force is set


bashbunni:
@aymanbagabas should I close this since `beta` is outdated?

PR #303, opened 2024-05-24
fix: properly follow the CLICOLOR standards

aymanbagabas:
See http://bixense.com/clicolors/


meowgorithm:
@aymanbagabas are we doing this in `v2`? If not, sounds like we might as well?

aymanbagabas:
> @aymanbagabas are we doing this in `v2`? If not, sounds like we might as well?

Actually this is now part of colorprofile and no longer needed in lipgloss. Closing...

PR #315, opened 2024-06-08
fix: Deprecate UnsetBorderTopBackgroundColor in favor of UnsetBorderTopBackground

nervo:
This looks like a looong standing typo :)

This is the only one method ending by "Color". To avoid breaking changes, i've just copy/paste to a new one, directly call this new one, and add a deprecation notice.


meowgorithm (approved):

meowgorithm:
Wow, indeed it was. @aymanbagabas, heads up for v2.

PR #154, opened 2022-10-31
Add support for gradient colours

schmorrison:
Adds a struct which implements TerminalColor interface. Added steps to Render for gradient application to text.

PR #325, opened 2024-06-24
feat: add Blend helper

lukechampine:
Based on https://github.com/charmbracelet/lipgloss/pull/154, this adds a simple helper for creating gradients between two colors, without changing anything about `Render`. I don't have strong opinions as to the API here, I'm just looking for an easy way to create nice gradients without having to import `termenv` separately.


meowgorithm:
Yeah, I think that makes sense. @lukechampine, want to open a PR over there?

I'd also suggest using [go-colourful](https://github.com/lucasb-eyer/go-colorful) directly (`termenv` is just wrapping it) and blending with the `HSLuv` algorithm.

lukechampine:
ah, thanks -- go-colorful has no dependencies, so that's perfect

PR #328, opened 2024-07-06
Fix README table example

frejonb:
I typically don't create PRs for trivial changes, but in this case I was learning `lipgloss` and it took me a bit to find the correct row styles from the table example in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,6 +410,21 @@ Lip Gloss ships with a table rendering sub-package.
 import "github.com/charmbracelet/lipgloss/table"
 ```
 
+Define the row styles.
+
+```go
+var (
+		// HeaderStyle is the lipgloss style used for the table headers.
+		HeaderStyle = re.NewStyle().Foreground(lipgloss.Color("99")).Bold(true).Align(lipgloss.Center)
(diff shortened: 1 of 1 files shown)


bashbunni:
Hey @frejonb thank you so much for your contribution! I also found table styling a bit confusing at first. I do think it makes sense to highlight this more explicitly for new users. I do have some feedback on this PR :)

If we're going to add information about styling lists it should look similar to the [list section](https://github.com/charmbracelet/lipgloss?tab=readme-ov-file#rendering-lists). Add a blurb mentioning custom styles, followed by a working example + capture of the output.

> Tables can have custom border styles set by `Lipgloss.Style` using the [`BorderStyle`](https://pkg.go.dev/github.com/charmbracelet/lipgloss@v0.12.1/table#Table.BorderStyle) function. Additionally, table content can be stylized using [`StyleFunc`](https://pkg.go.dev/github.com/charmbracelet/lipgloss@v0.12.1/table#StyleFunc) which supports conditional formatting.

```go
// an example using BorderStyle and StyleFunc
```

<an image generated with `freeze -x "go run ./my-example" -r 10 --window --font.size 36 --margin 20`>

PR #329, opened 2024-07-10
Initializing t.data with new content instead of leaving it as nil

zveinn:
leaving t.data as nil causes a panic if the table is re-rendered again without adding new rows

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -84,7 +84,7 @@ func New() *Table {
 
 // ClearRows clears the table rows.
 func (t *Table) ClearRows() *Table {
-	t.data = nil
+	t.data = NewStringData()
 	return t
 }
 


aymanbagabas:
@zveinn Could you please add a test case?

zveinn:
Not really, I´m too busy to write that for you :S sorry ..

siennathesane:
Here's the test code to be added...

```go
func TestEmptyTable(t *testing.T) {
	table := New().
		StyleFunc(TableStyle).
		Headers("Fruit", "Color", "Code")

	expected := strings.TrimSpace(`
┌───────┬───────┬──────┐
│ Fruit │ Color │ Code │
├───────┼───────┼──────┤
└───────┴───────┴──────┘
`)

	if stripString(table.String()) != expected {
		t.Fatalf("expected:\n\n%s\n\ngot:\n\n%s", expected, stripString(table.String()))
	}
}
```

bashbunni:
I believe this is fixed in https://github.com/charmbracelet/lipgloss/pull/372

PR #338, opened 2024-07-18
feat: add Title option for lists

bashbunni:


bashbunni:
This is needed to fully replace the features of `tree` where we were able to set a name for the root node.

bashbunni:
- [ ] support `any` type instead of just string

bashbunni:
Merging with https://github.com/charmbracelet/lipgloss/pull/339

PR #339, opened 2024-07-18
feat: use tree enumerators

bashbunni:
- **feat: add Root option for lists**
- **refactor: use tree enumerators**

When a list is used as the root of a list, it merges the items and inherits the name of the root list. Otherwise it looks like the sub-list belongs to the last element in the root list.

e.g. 
image
> This but imagine the "More Documents", etc are indented

Instead it looks like
image

Where the root tree's items are appended to the end of the list


bashbunni:
Builds off of the other PR that adds the `Title` func to set a name for the list

meowgorithm:
Looking good. Want to merge this with #339 so we can treat this as a general pass over `tree`/`list`? I think it'll be easier to reason about that way.

caarlos0 (approved):

bashbunni:
I think we should have the `Indenter` default to `TreeIndenter` when we set the Enumerator to `list.Tree` or `list.RoundedTree`. It feels weird having to specify that every time

maaslalani:
Agreed!

bashbunni:
🌴

bashbunni:
Closing in favour of bringing trees out of internal https://github.com/charmbracelet/lipgloss/pull/342

PR #340, opened 2024-07-22
fix: table truncation

caarlos0:
it was truncating "too late", causing style leaks

refs https://github.com/charmbracelet/glamour/pull/317

### DIFF
--- a/table/table_test.go
+++ b/table/table_test.go
@@ -694,13 +694,13 @@ func TestTableWidthShrink(t *testing.T) {
 
 	expected := strings.TrimSpace(`
 ┌─────────┬─────────┬────────┐
-│ LANGUAG │ FORMAL  │ INFORM │
+│ LANGUA… │ FORMAL  │ INFOR… │
 ├─────────┼─────────┼────────┤
 │ Chinese │ Nǐn hǎo │ Nǐ hǎo │
 │ French  │ Bonjour │ Salut  │
-│ Japanes │ こんに  │ やあ   │
-│ Russian │ Zdravst │ Privet │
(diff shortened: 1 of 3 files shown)


table/table_test.go
```diff
@@ -725,7 +725,7 @@ func TestTableWidthSmartCrop(t *testing.T) {
 
 	expected := strings.TrimSpace(`
 ┌──────┬─────┬──────────┐
-│ Name │ Age │ Location │
```
maaslalani, line 728: I believe this is incorrect. It should be `Age` not `Ag…`

table/table_test.go
```diff
@@ -757,15 +757,16 @@ func TestTableWidthSmartCropExtensive(t *testing.T) {
...
 ┣━━━━╋━━━━━╋━━━━━┫
-┃ Ch ┃ 您  ┃ 你  ┃
-┃ Ja ┃ こ  ┃ や  ┃
-┃ Ar ┃ أهل ┃ أهل ┃
-┃ Ru ┃ Здр ┃ При ┃
-┃ Sp ┃ Hol ┃ ¿Qu ┃
-┃ En ┃ You ┃ How ┃
+┃ C… ┃ 您… ┃ 你… ┃
+┃ J… ┃ こ… ┃ や… ┃
+┃ A… ┃ أه… ┃ أه… ┃
```
maaslalani, line 764: I believe this line is incorrect.

maaslalani:
I might be incorrect but I think I decision to not prioritize padding over text content. This is important when resizing, so that we remove padding before truncating content.

caarlos0:
yeah this was actually another issue

PR #342, opened 2024-07-23
feat: move tree to root

caarlos0:
closes #339 
closes #341


bashbunni (approved):

aymanbagabas (approved):

meowgorithm:
If you don't mind, going to look at a few last things today before we merge this one.

bashbunni:
Hey @meowgorithm any changes on this one?

meowgorithm:
Let’s get the tree examples back in as well — I'm looking at stuff now.

bashbunni:
@meowgorithm kk no rush. Just added the examples back

meowgorithm (approved):
Feeling good! Let's get this done ⚡️

bashbunni:
Removed `JoinVertical` from the final rendered element as it was writing unstyled whitespace to the final string. It does that by design to appropriately align elements in the block, but because it doesn't have access to the style, the whitespace can't be rendered. That said, I haven't seen this issue outside of the tree examples, so it's not pressing. 

If we did want to fix this we could replace this whitespace rendering with calling `Align` on the style which can render the whitespace properly or fix this with our compositor in a future version of lipgloss

bashbunni (changes requested):
TODO before merge
- [ ] tidy godoc (some tree docs mention list)
- [x] show an example of unique styling of the root element (this might be broken)
- [ ] add tree info to `README.md`

bashbunni:
I thought the lipgloss root styling might be broken because if we do 

```go
ItemStyleFunc(func(children tree.Children, i int) lipgloss.Style {
    if children.At(i).Value() == "Nyx" {                          
        return itemStyle.Background(lipgloss.Color("#04B575"))    
    }                                                             
    return itemStyle                                              
})                                                                
```
we get 
```
%!v(PANIC=String method: runtime error: invalid memory address or nil pointer dereference)
``` 
because when we're on the root item, it's `children.At(-1)` which returns `nil`, then we attempt to call `Value()` on it. 

_As-is we will always need to do a root value check in `ItemStyleFunc` to avoid this..._

bashbunni:
We might even want to see about separating out `RootStyle` from `ItemStyle` 💭

caarlos0:
> We might even want to see about separating out `RootStyle` from `ItemStyle` 💭

i think that makes more sense yes

README.md
```diff
@@ -562,6 +562,104 @@ for i := 0; i < repeat; i++ {
...
+// .
+// ├── A
+// ├── B
+// └── C
+```
+
+Trees have the ability to nest.
+
+```go
+t := tree.New().
```
bashbunni, line 595: I can use the shorthand `tree.Root(".")...` and get rid of `New()` (not needed)

bashbunni (approved):

aymanbagabas (approved):

meowgorithm (approved):

Issue #356, opened 2024-08-28
setting table height doesn't do anything


bashbunni:
We should cut off additional rows when a height is explicitly set. The last allowed row should probably just show '...' to show that there's more content than the height allows. Checked the source code and the resizing logic is only applied to width

code to reproduce:
```go
package main

import (
	"fmt"

	"github.com/charmbracelet/lipgloss"

	glossytable "github.com/charmbracelet/lipgloss/table"
)

func main() {
	gloss := glossytable.New().
		Headers("ID", "title", "description").
		Row("1", "Lost in Time", "A thrilling adventure through the ages.").
		Row("2", "Whispering Shadows", "Secrets unravel in a haunted town.").
		Row("3", "Stolen Identity", "An innocent man's life turned upside down.").
		Row("4", "Into the Abyss", "Exposing deep-rooted conspiracies.").
		Border(lipgloss.NormalBorder()).
		Width(20).
		Height(4)

	fmt.Println(gloss)
}
```

Broderick-Westrope:
@bashbunni can you clarify what is meant to impact the height of the table?
1. Just the number of lines?
2. Or the number of rows?
3. Or the rows + header?

To demonstrate, the following is a table of height 4 using each of these approaches:
```
1.
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
╰──┴───────────────┴───────────────────╯

2.
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
│1 │Lost in Time   │A thrilling advent…│
│2 │Whispering Sha…│Secrets unravel in…│
│3 │Stolen Identity│An innocent man's …│
│4 │Into the Abyss │Exposing deep-root…│
╰──┴───────────────┴───────────────────╯

3.
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
│1 │Lost in Time   │A thrilling advent…│
│2 │Whispering Sha…│Secrets unravel in…│
│3 │Stolen Identity│An innocent man's …│
╰──┴───────────────┴───────────────────╯
```

NOTE: This example isn't concerned with adding ellipsis as the bottom row ;)

Currently `bubbles/table` takes a hybrid approach where the height is the sum of two things:
- `lipgloss.Height(headers)` which takes into account styling on the headers (eg. borders).
- and the length of the rows slice without styling.

[...]

PR #358, opened 2024-08-29
fix(table): use table height

Broderick-Westrope:
Fixes https://github.com/charmbracelet/lipgloss/issues/356

## Changes
* `table/table.go`: Modified the `constructRow` method to accept an `isOverflow` parameter, allowing rows to be rendered as overflow rows with ellipsis when necessary.
* `table/table.go`: Updated the `String` method to calculate the available height for rendering rows and handle the rendering of overflow rows correctly. Most of this lives in the new `constructRows` method.
* `table/table_test.go`: Added unit tests for the `Height` method.
* `table/table.go`: Added a boolean `useManualHeight` to the `Table` struct. I thought it would be good to maintain the existing functionality (auto table height) by default. I've put this in one commit so it can be easily removed if needed. The `useManualHeight` boolean tracks whether the user has manually set the height of the table. When `false`, the new height functionality will not be used. It is `false` by default and only set to `true` when the height is set (`table.Height(x)`).
  * Without this the existing table unit tests (and likely other users' code) would would have to change every `New` call to have a corresponding `Height` call. With this feature, height is flexible and calculated for you by default.
  * We could replace this bool with an understanding that `height == -1` represents auto height. Not sure what is preferred.

### How Height is Used

The default behaviour is the same as before, now there's just a few more features:
- Calling `New` without calling `Height` will render the full table (despite the `Table.height` value being 0). Example: the following table _technically_ has a height of 0, but the user did not set that so we assume they want to show the whole table:
```
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
│1 │Lost in Time   │A thrilling advent…│
│2 │Whispering Sha…│Secrets unravel in…│
│3 │Stolen Identity│An innocent man's …│
│4 │Into the Abyss │Exposing deep-root…│
╰──┴───────────────┴───────────────────╯
```
_NOTE: When the following examples say "set a height" they mean make a call to the `table.Height` method_
- If you set a height larger than the tables total height, it won't look any different. No extra lines are added. This is the same if you choose the exact height needed. Example, we could set the table height to 8 or 800 and we would still get this:
```
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
│1 │Lost in Time   │A thrilling advent…│
│2 │Whispering Sha…│Secrets unravel in…│
│3 │Stolen Identity│An innocent man's …│
│4 │Into the Abyss │Exposing deep-root…│
╰──┴───────────────┴───────────────────╯
```
- If you set a height so small that no rows can be rendered, and overflow row will still be rendered. Example, we could set the table height to 5 or anything less (including negatives) and we would get:
```
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
│… │...            │...                │
╰──┴───────────────┴───────────────────╯
```
- And finally, if you set a height between the last two examples (ie. enough to show a row but not enough to show all rows) then we render as many as possible finishing with an overflow row. Example, the following table has a height of 6:
```
╭──┬───────────────┬───────────────────╮
│ID│title          │description        │
├──┼───────────────┼───────────────────┤
│1 │Lost in Time   │A thrilling advent…│
│… │...            │...                │
╰──┴───────────────┴───────────────────╯
```

## Notes
- The height of a table is defined as the number of lines (ie. vertical cells) that the table takes up. This includes headers, data rows, and any styling.
- Since at least one data row is **always** displayed (even if it is just an overflow row), if the user sets the height to anything less than 5 but there is only one row of data to show, the data row will be shown instead of an overflow row. This stems from an assumption that if we are going to "force print" 5 lines (aka. 5 vertical cells) of text and there is only one data row anyway, we might as well show the user some useful information rather than an empty or overflow row.
- If the only row printed is an overflow row, the column widths are calculated for the original data and yet we are only printing ellipsis. Let me know if this is something that needs to be altered.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -532,14 +532,9 @@ func (t *Table) constructRow(index int, isOverflow bool) string {
 	for c := 0; c < t.data.Columns(); c++ {
 		cellWidth := t.widths[c]
 
-		// TODO(fix): small cells should be using '…' instead of '...'. This is rendered wrong when
-		// 	then cellWidth is greater than 3 but the visible cell (ie. after styling) is smaller than 3.
-		cell := "..."
-		switch {
-		case !isOverflow:
+		cell := "…"
+		if !isOverflow {
 			cell = t.data.At(index, c)
-		case cellWidth < 3:
-			cell = "…"
 		}
 
 		cells = append(cells, t.style(index+1, c).
--- a/table/table.go
+++ b/table/table.go
@@ -359,20 +359,20 @@ func (t *Table) String() string {
 
 	// If there are no data rows render nothing.
 	if t.data.Rows() > 0 {
-		topHeight := lipgloss.Height(sb.String()) - 1 // Subtract 1 for the newline.
+		// The height of the top border. Subtract 1 for the newline.
+		topHeight := lipgloss.Height(sb.String()) - 1
+
 		availableHeight := t.height - (topHeight + lipgloss.Height(bottom))
 		rowsToRender := min(availableHeight, t.data.Rows())
 
-		// The following replicates a do-while loop. When there is data we always render at least one row.
-		// Whenever the height is too small to render all rows, the bottom row will be an overflow row (ellipsis).
-		rowIdx := t.offset
-		for ok := true; ok; ok = rowIdx < rowsToRender {
-			isOverflow := rowsToRender == 0 ||
-				(rowIdx == rowsToRender-1 && rowsToRender < t.data.Rows())
+		// When there is data we always render at least one row.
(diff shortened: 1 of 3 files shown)


Broderick-Westrope:
@bashbunni hoping to close this one out. Any feedback or blockers?

bashbunni (approved):
Hey this looks great. Thank you so much for the very thorough work and tests! This was really quick to review as a result 🙏

PR #359, opened 2024-08-30
(v2) feat: add adaptive color package

aymanbagabas:
This introduces a helper type `LightDark` that takes a boolean to determine which `Color(light, dark)` to choose from. The `adaptive` package is a helper package that uses the `lipgloss.LightDark` along with querying the terminal when the module is imported to choose the appropriate light-dark color.

This will also make `lipgloss.Color` take `any` type and try to infer the color from either strings or integers.

Example:

```go
var (
  light = "#0000ff"
  dark = "#ff0000"
)

colorToUse := adaptive.Color(light, dark) // the terminal is queried before choosing the color
fmt.Println(colorToUse)
```


color.go
```diff
@@ -107,6 +112,27 @@ func (c RGBColor) RGBA() (r, g, b, a uint32) {
...
+// used based on the terminal's background color.
+//
+// Example usage:
+//
+//	useDark := lipgloss.LightDark(true)
+//	light := "#0000ff"
+//	dark := "#ff0000"
+//	colorToUse := useDark.Color(light, dark)
+//	fmt.Println(colorToUse)
+type LightDark bool
```
bashbunni, line 125: What do you think about naming this `Adapt` instead? Might feel more familiar.

Then it could be 
```
adaptive := lipgloss.Adapt(true)
colors := adaptive.Color("#0000ff", "#ff0000")
```

I think that would make the purpose more clear

meowgorithm: Totally: this is really nice.

meowgorithm: I would also update the comment to show use with Bubble Tea, as well as standalone Lip Gloss (when we get there).

meowgorithm (changes requested):

color.go
aymanbagabas: I'm thinking of dropping this type which is basically an ifelse; the adaptive sub-package should be enough imo

PR #372, opened 2024-09-20
fix(table): unset data rows without causing nil pointer err

bashbunni:
Without this change `data` is set to `nil` causing nil ptr errors whenever we want to add data to the table after calling `ClearRow`. Instead, this function should reset the values without clearing the data type so it's still usable

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -85,7 +85,7 @@ func New() *Table {
 
 // ClearRows clears the table rows.
 func (t *Table) ClearRows() *Table {
-	t.data = nil
+	t.data = NewStringData()
 	return t
 }
 


meowgorithm:
Okay cool. So basically if you call `ClearRows()` and then try and add more data we get a nil pointer exception, right?

```go
table := New().
		Headers("LANGUAGE", "FORMAL", "INFORMAL").
		Row("Chinese", "Nǐn hǎo", "Nǐ hǎo")
	table.ClearRows() // Right here
	table.Row("French", "Bonjour", "Salut")
```

bashbunni:
@meowgorithm yes, exactly! Or when you would call `String()` it would try and check the length of rows and get a nil pointer error because the value is nil instead of an array or slice

caarlos0 (approved):

meowgorithm (approved):

PR #373, opened 2024-09-20
fix(table): do not shrink table with offset

bashbunni:
We should fix this before merging. This failing test is to provide context on this issue and how to reproduce it. I believe in this case we should honour the height rather than the offset since we are more likely to see dependents on the height.


bashbunni:
Fixed the issue in the second commit :)

table/table_test.go
```diff
@@ -1140,6 +1140,120 @@ func TestTableHeightWithOffset(t *testing.T) {
...
+		{"98", "Rome", "Italy", "4,297,877"},
+		{"99", "Shijiazhuang", "China", "4,285,135"},
+		{"100", "Montreal", "Canada", "4,276,526"},
+	}
+	table := New().
+		Rows(rows...).
+		Offset(80).
+		Height(45)
+
+	got := lipgloss.Height(table.String())
```
bashbunni, line 1251: TODO clean this up a bit...

meowgorithm:
Hey Bash! So what's the bug you're solving here exactly?

bashbunni:
Hey, sorry I thought I has included screenshots or something. When you get to a certain offset, it would shrink the table instead of just moving the cursor, causing the height of the table to get smaller than the height specified by the user. If you run the test on master vs this branch you'll be able to see what I mean. Sorry I'll include images/recordings next time :)

meowgorithm:
Wow, what a bug! Anyway, the fix looks good so far. To state the issue simply:

> When the table is rendered with an offset the height of the table doesn't always match the actual height that was set.

Does that sound correct?

Here is some code to reproduce the issue:
https://gist.github.com/meowgorithm/99a20cd66739a3d0550213c06da0f9a1

Let's make sure we test this on some real-world examples, including with the [Table Bubble](https://github.com/charmbracelet/bubbles?tab=readme-ov-file#table).

bashbunni:
@meowgorithm yeah that's right! The table bubble doesn't currently use lipgloss table, that's how I discovered this bug was through porting the table bubble to use lipgloss' table renderer. As soon as there was the interactivity element, it broke. 

This is the table bubble wip - https://github.com/charmbracelet/bubbles/pull/617 that you can use with this branch. I can make this easier to test/demo over the next few days :)

meowgorithm:
Aha! That's right, `table` in Bubbles is totally its own thing. Okay cool, I'll test a bit more but generally this looks good.

meowgorithm (approved):
Alright, LGTM. Such a good fix!!

PR #377, opened 2024-09-24
fix: shared indices for first row of data and headers (StyleFunc bug)

bashbunni:
While migrating the table bubble to use Lip Gloss, I noticed some rendering issues where the cursor was constantly off by one. After fixing the off by one error, I was still having some range issues where the first element was never showing the selected style. Turns out, both the first element in the list and the header are the same index, so whatever style was returned first would be rendered by both the header and first element in the list.

Despite the cursor being in the right position, it was rendering the _next_ element in the list with the declared styles.

> [!NOTE]
> This is a _bit_ of a breaking change, though necessary for the first element to ever render correctly when there are also styles set on the headers. This would require table users to update their `StyleFunc` implementations to use `HeaderRow` instead of 0.

Before:
image

After:
image
(it's failing intentionally so I could see the outputs)

This PR introduces a test to Lip Gloss that adds unique styles to both the header and first row of data to showcase the issue. When running the test against `master`, you'd need to comment out the `HeaderRow` stuff and you'll see that both the first row and header use the same style.


bashbunni:
This is the test that reproduces the behaviour I was seeing in the bubble (in the screenshots above) https://github.com/charmbracelet/lipgloss/blob/fix-stylefunc/table/table_test.go#L1144-L1169

table/table.go
```diff
@@ -7,6 +7,10 @@ import (
 	"github.com/charmbracelet/x/ansi"
 )
 
+// HeaderRow denotes the header's row index used when rendering headers. Use
+// this value when looking to customize header styles in StyleFunc.
+const HeaderRow int = -1
```
meowgorithm, line 12: Does this need to be exported?

meowgorithm:
Thanks, @bashbunni! If it's not too much trouble, can you plop down some source code for a quick program to reproduce this (even if it's based on the tests)? It's helpful to see the before/after interactively to understand what we're lookin' for.

table/table.go
bashbunni: Yeah this is the value they would be using in `StyleFunc` to style the header instead of 0. I figured better to export a const than have them check for -1

bashbunni:
@meowgorithm the tests are probably the best place to see the changes!

caarlos0 (approved):

PR #384, opened 2024-10-03
feat: Add border text functions for top and bottom borders.

pachecot:
This allows to set titles and status information in the borders.

This solves the issue where I keep trying to add some status information to the bottom of a border. I either need to take it apart and rebuild it with the additional information or create my own bottom border. This also allows setting some title information, which is nice too. 

this still need some error checking if someone returns longer strings. 

```go
type BorderFunc func(width int, middle string) string

func (s Style) BorderTopFunc(p Position, bf BorderFunc) Style 
func (s Style) BorderBottomFunc(p Position, bf BorderFunc) Style 
```

Example.. 
```go
    NewStyle().
        Width(10).
        Border(NormalBorder()).
        BorderTopFunc(Left, func(width int, middle string) string {
            return "TITLE" 
        })
```

```text
┌TITLE─────┐
│          │
└──────────┘
```

will allow to set functions for each corner 

```text
┌A───B───C┐
│         │
└D───E───F┘
```

### DIFF
--- a/borders.go
+++ b/borders.go
@@ -26,6 +26,8 @@ type Border struct {
 	MiddleBottom string
 }
 
+type BorderFunc func(width int, middle string) string
+
 // GetTopSize returns the width of the top border. If borders contain runes of
 // varying widths, the widest rune is returned. If no border exists on the top
 // edge, 0 is returned.
@@ -248,6 +250,9 @@ func (s Style) applyBorder(str string) string {
 		rightBG  = s.getAsColor(borderRightBackgroundKey)
 		bottomBG = s.getAsColor(borderBottomBackgroundKey)
 		leftBG   = s.getAsColor(borderLeftBackgroundKey)
+
+		topFuncs    = s.borderTopFunc
+		bottomFuncs = s.borderBottomFunc
 	)
 
 	// If a border is set and no sides have been specifically turned on or off
@@ -327,7 +332,12 @@ func (s Style) applyBorder(str string) string {
 
 	// Render top
 	if hasTop {
-		top := renderHorizontalEdge(border.TopLeft, border.Top, border.TopRight, width)
+		var top string
+		if len(topFuncs) > 0 {
+			top = renderAnnotatedHorizontalEdge(border.TopLeft, border.Top, border.TopRight, topFuncs, width)
+		} else {
+			top = renderHorizontalEdge(border.TopLeft, border.Top, border.TopRight, width)
+		}
 		top = s.styleBorder(top, topFG, topBG)
 		out.WriteString(top)
 		out.WriteRune('\n')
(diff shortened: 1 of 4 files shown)


pachecot:
made some changes to names and api so will accept strings and func() string.

here is sample..
```
	m := struct {
		index  int
		count  int
		status string
	}{
		index:  5,
		count:  10,
		status: ":)",
	}

	reverseStyle := lipgloss.NewStyle().Reverse(true)

	t := lipgloss.NewStyle().
		Width(40).
		Height(10).
		BorderDecoration(lipgloss.NewBorderDecoration(
			lipgloss.Top,
			lipgloss.Center,
			reverseStyle.Padding(0, 1).Render("BIG TITLE"),
		)).
		BorderDecoration(lipgloss.NewBorderDecoration(
			lipgloss.Bottom,
			lipgloss.Right,
			func(width int, middle string) string {
				return reverseStyle.Render(fmt.Sprintf("[%d/%d]", m.index+1, m.count)) + middle
			},
		)).
		BorderDecoration(lipgloss.NewBorderDecoration(
			lipgloss.Bottom,
			lipgloss.Left,
			reverseStyle.Padding(0, 1).SetString(fmt.Sprintf("Status: %s", m.status)).String,
		))

	fmt.Println(t)
```

image

meowgorithm:
Very nice! We had originally resigned to not including this sort of functionality, but your implementation looks really good. A couple open questions:

* What should happen when a label is wider than the width of the box? Truncate? Expand the box?
* What should happen when two labels collide? For example, when we have long left and right labels.

Can you also fix the issues pointed out with the linter?

pachecot:
thanks, 

I will address the issues in the linter. It looks like a bug.

It is intended to limit to the length of the box truncating the texts. Currently, it is first come first served, Left -> Center -> Right. so it is unfair and some may not even be included. I would like to change it to a more fair truncating, where they all get some of the space.

Currently they could collide. that will need to be addressed.

pachecot:
Made some changes to equalize the sizes some and provide space between the labels. Currently does not add ellipsis, may want to add them. 

```
t := style: NewStyle().
    Width(16).
    Border(NormalBorder()).
    BorderDecoration(NewBorderDecoration(Top, Left, "LeftLeftLeftLeft")).
    BorderDecoration(NewBorderDecoration(Top, Center, "CenterCenterCenter")).
    BorderDecoration(NewBorderDecoration(Top, Right, "RightRightRightRight"))
```

```
┌LeftL─Cent─Right┐
│                │
└────────────────┘
```

pachecot:
I figured out how to include side borders also, so made some additional updates.

The support for style for the side borders are limited. I am not sure if there are existing functions that could support this better.

image

matthjes:
This would be really helpful. Is there any intention to merge this?

PR #386, opened 2024-10-07
fix(render): strip carriage returns from strings

bashbunni:
If a string contains carriage returns `\r`, it will cause some funky issues when rendering. More specifically any time we do any string concatenation, string A will be overwritten by string B instead of appended.

this is an example log output when a string has a `\r` and goes through `alignHorizontal` to pad spaces 
```
2024/10/07 13:36:43 end of horizontal align
git: 'defer' is not a git command. See 'git --help'.
----------------------------------------------------
-----------------------ds are
------------------------------------------
------------------------------------------
----------------------------------------------------
```

After this change:
```
2024/10/07 13:53:24 end of horizontal align
git: 'defer' is not a git command. See 'git --help'.
----------------------------------------------------
The most similar commands are-----------------------
    rerere------------------------------------------
    revert------------------------------------------
----------------------------------------------------
```

related:
https://github.com/charmbracelet/freeze/pull/119

### DIFF
--- a/style.go
+++ b/style.go
@@ -353,6 +353,8 @@ func (s Style) Render(strs ...string) string {
 
 	// Potentially convert tabs to spaces
 	str = s.maybeConvertTabs(str)
+	// carriage returns can cause strange behaviour when rendering.
+	str = strings.ReplaceAll(str, "\r", "")
 
 	// Strip newlines in single line mode
 	if inline {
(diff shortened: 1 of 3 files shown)


bashbunni:
Added a test that reproduces this behaviour. I notice that it only happens when `Margin` is set which is interesting. If you want to see for yourself, run the test on `master` and you'll see it fail

style.go
```diff
@@ -353,6 +353,8 @@ func (s Style) Render(strs ...string) string {
 
 	// Potentially convert tabs to spaces
 	str = s.maybeConvertTabs(str)
+	// carriage returns can cause strange behaviour when rendering.
+	str = strings.ReplaceAll(str, "\r", "")
```
aymanbagabas, line 357:
```suggestion
	str = strings.ReplaceAll(str, "\r\n", "\n")
```

PR #387, opened 2024-10-08
(v2) chore: update tests and remove unused code

aymanbagabas:


style_test.go
```diff
@@ -15,29 +15,29 @@ func TestUnderline(t *testing.T) {
...
+			"\x1b[4;4ma\x1b[m\x1b[4;4mb\x1b[m\x1b[4m \x1b[m\x1b[4;4mc\x1b[m",
 		},
 		{
 			NewStyle().Underline(true).UnderlineSpaces(true),
-			"\x1b[4;4ma\x1b[0m\x1b[4;4mb\x1b[0m\x1b[4m \x1b[0m\x1b[4;4mc\x1b[0m",
+			"\x1b[4;4ma\x1b[m\x1b[4;4mb\x1b[m\x1b[4m \x1b[m\x1b[4;4mc\x1b[m",
 		},
 		{
 			NewStyle().Underline(true).UnderlineSpaces(false),
-			"\x1b[4;4ma\x1b[0m\x1b[4;4mb\x1b[0m \x1b[4;4mc\x1b[0m",
```
bashbunni, line 26: why remove the reset?

style_test.go
```diff
@@ -119,59 +115,9 @@ func TestStyleRender(t *testing.T) {
...
 		res := s.Render()
 		if res != tc.expected {
-			t.Errorf("Test %d, expected:\n\n`%s`\n`%s`\n\nActual output:\n\n`%s`\n`%s`\n\n",
-				i, tc.expected, formatEscapes(tc.expected),
-				res, formatEscapes(res))
-		}
-	}
-}
-
-func TestStyleCustomRender(t *testing.T) {
```
bashbunni, line 129: why remove this?

tree/testdata/TestRootStyle.golden
```diff
@@ -1,3 +1,3 @@
-[48;2;89;86;224mRoot[0m
```
bashbunni, line 1: We need to keep ansi sequences in this golden file since we're testing that the root element styling is different from the other elements in the list. We also need to remove the `ansi.Strip` from the `TestRootStyle`

tree/tree_test.go
```diff
@@ -413,7 +413,6 @@ Root
 }
 
 func TestRootStyle(t *testing.T) {
-	lipgloss.SetColorProfile(termenv.TrueColor)
```
bashbunni, line 416: We do want this test to output truecolor output. Since v2 always outputs colors this makes sense, but we shouldn't strip the ansi from the output (part of why I'm using golden file so it retains the ansi sequence)

bashbunni (changes requested):
We need to keep the colorized output for `TestRootStyle`

style_test.go
aymanbagabas: We're not removing them, `\x1b[0m` is the same as `\x1b[m`. One less byte to reset the colors.

style_test.go
aymanbagabas: We no longer have "renderers"

style_test.go
bashbunni: ouuuuuu ok thank you!!

meowgorithm:
I believe all the work here found its way to `v2-exp` via #397. Let me know if it's okay to close and delete this branch @aymanbagabas and @bashbunni.

aymanbagabas:
> I believe all the work here found its way to `v2-exp` via #397. Let me know if it's okay to close and delete this branch @aymanbagabas and @bashbunni.

That's great! Let's close and delete then

bashbunni:
Sounds good!

PR #389, opened 2024-10-10
(v2) fix(examples): make examples use v2

bashbunni:
Goal: get the examples working for v2

- [x] ssh - tested with ANSI256 & truecolor and it's giving the right output
- [x] table examples
- [x] tree
- [x] layout
- [x] list

Only thing outstanding is clarifying how to work with light and dark backgrounds + detecting background color with v2. On that note, https://github.com/charmbracelet/lipgloss/pull/359 is related to this PR


bashbunni:
I also had to update log @ `v2-exp` to use `colorprofile`'s Writer. Let me know the best way to navigate the un-merged deps. In this case I just have a `go mod replace` directive to point it to a local version of log checked out at `v2-exp` :)

meowgorithm:
@bashbunni so the work here has been merged into `v2-adaptive-standalone`, which contains solutions we're working though for adaptive colors, and additional fixes for the open items here that were (excellently) documented.

Are you cool if I open a PR for that branch and supersede this one?

bashbunni:
@meowgorithm yeah totally cool! If we don't need this branch anymore we can close it

Issue #383, opened 2024-09-28
table resizing issue with \r\n in cells


pachecot:
something in the resize logic seems to fail on a carriage return line feed "\r\n"

renders ok without setting the Width

```
package main

import (
	"fmt"
	"os"

	tea "github.com/charmbracelet/bubbletea"
	"github.com/charmbracelet/lipgloss"
	"github.com/charmbracelet/lipgloss/table"
)

var (
	data = [][]string{
		{"a0", "b0", "c0", "d0"},
		{"a1", "b1.0\r\nb1.1\r\nb1.2\r\nb1.3\r\nb1.4\r\nb1.5\r\nb1.6", "c1", "d1"},
		{"a2", "b2", "c2", "d2"},
		{"a3", "b3", "c3", "d3"},
	}
)

type Model struct {
	table *table.Table
}

func InitialModel() *Model {
	return &Model{
		table: table.New().
			Border(lipgloss.NormalBorder()).
			Data(table.NewStringData(data...)),
	}
}

func (m *Model) Init() tea.Cmd {
	return nil
}

func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {

	switch msg := msg.(type) {

	case tea.WindowSizeMsg:
		m.table.Width(msg.Width)

	case tea.KeyMsg:
		switch msg.String() {
		case "q", "ctrl+c":
			return m, tea.Quit
		}
	}
	return m, nil
}

func (m *Model) View() string {
	return m.table.String()
}

func main() {
	p := tea.NewProgram(
		InitialModel(),
		tea.WithAltScreen(),
	)
	if _, err := p.Run(); err != nil {
		fmt.Printf("Alas, there's been an error: %v", err)
		os.Exit(1)
	}
}
```

this is the result I got
```
┌──────────────────┬────────────────────┬──────────────────┬──────────────────┐
│a0                │b0                  │c0                │d0                │
                │c1                │d1                │
                │                  │                  │
                │                  │                  │
                │                  │                  │
                │                  │                  │
                │                  │                  │
│                  │b1.6                │                  │                  │
│a2                │b2                  │c2                │d2                │
│a3                │b3                  │c3                │d3                │
```

edit:

pachecot:
Seems like issue is in x/ansi.Wrap isn't trimming the '\r' from "\r\n" and adding spaces so getting stuff like "\r    \n".

bashbunni:
Added a test for this case in the linked PR :)

[...]

PR #390, opened 2024-10-11
fix(getLines): carriage return was not being removed with new lines.

pachecot:
getLines function is returning lines with out trimming the trailing carriage return.

anything using this might be appending strings to the carriage return terminated strings causing formatting issues.

Related 
#383 
#386


bashbunni:
Quick note on this one: we should probably follow a similar implementation to what we did [here](https://github.com/charmbracelet/lipgloss/blob/022e96717265bb4650a770f5abf2b3c5fcc85e58/style.go#L357) using `strings.ReplaceAll` to strip all `\r` from the string 💭

PR #393, opened 2024-10-11
feat: simplify adaptive package and expose QueryBackgroundColor

aymanbagabas:
Remove unnecessary global variables and functions from the adaptive package. Instead of using a global variable to store the color function, the color function is now created on the fly when needed.

### DIFF
--- a/adaptive/adaptive.go
+++ b/adaptive/adaptive.go
@@ -1,47 +1,38 @@
 package adaptive
 
 import (
+	"image/color"
 	"os"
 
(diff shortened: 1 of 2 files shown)


bashbunni:
Do we still need this now that we have the `compat` package?

bashbunni:
Do we still need this since we have the `compat` package?

meowgorithm:
Superseded by the `compat` package; closing accordingly.

PR #405, opened 2024-10-18
chore(lint): update linter

bashbunni:
Uses the [meta repo `lint.yml`](https://github.com/charmbracelet/meta/blob/main/.github/workflows/lint.yml). Deleted the `lint-soft` workflow as we don't have one declared in the meta repo


bashbunni:
Ah ok, thank you! Just added a commit to remove them. Note that in the `colorprofile` repo, we still have `golangci` files too. Not sure if we still need them https://github.com/charmbracelet/colorprofile/blob/main/.golangci.yml

aymanbagabas:
> Ah ok, thank you! Just added a commit to remove them. Note that in the `colorprofile` repo, we still have `golangci` files too. Not sure if we still need them https://github.com/charmbracelet/colorprofile/blob/main/.golangci.yml

Thanks for noticing that, we don't need those files anymore as the shared workflow uses the golangci files from the meta repository

aymanbagabas (approved):

PR #411, opened 2024-10-24
fix: border size getters when implicit borders are present

meowgorithm:
When a border style is set and no borders have been explicitly enabled or disabled, a border is implicitly applied to all sides of the block. Prior to this fix border size getters would incorrectly return 0 when an implicit border is set. This revision corrects that behavior.

In other words:

```go
s := lipgloss.NewStyle.Border(lipgloss.NormalBorder)

// Before
fmt.Println(s.GetBorderLeftSize()) // 0

// After
fmt.Println(s.GetBorderLeftSize()) // 1
```

### DIFF
--- a/get.go
+++ b/get.go
@@ -300,7 +300,7 @@ func (s Style) GetBorderTopWidth() int {
 // runes of varying widths, the widest rune is returned. If no border exists on
 // the top edge, 0 is returned.
 func (s Style) GetBorderTopSize() int {
-	if !s.getAsBool(borderTopKey, false) {
+	if !s.getAsBool(borderTopKey, false) && !s.implicitBorders() {
 		return 0
 	}
 	return s.getBorderStyle().GetTopSize()
@@ -310,7 +310,7 @@ func (s Style) GetBorderTopSize() int {
 // runes of varying widths, the widest rune is returned. If no border exists on
 // the left edge, 0 is returned.
 func (s Style) GetBorderLeftSize() int {
-	if !s.getAsBool(borderLeftKey, false) {
+	if !s.getAsBool(borderLeftKey, false) && !s.implicitBorders() {
 		return 0
 	}
 	return s.getBorderStyle().GetLeftSize()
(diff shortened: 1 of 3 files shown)


meowgorithm:
@bashbunni Just a heads up about this one. Technically it's a behavioral change but I think it _is_ the right approach here. I plan on merging it today.

I encountered the need for the fix in https://github.com/charmbracelet/bubbletea/pull/1204.

bashbunni:
Hey @meowgorithm just a heads up, this is super similar to another open PR https://github.com/charmbracelet/lipgloss/pull/282/files

The main difference being naming + I like your `GetBorderXSize()` implementations better.

get.go
```diff
@@ -519,6 +519,23 @@ func (s Style) getBorderStyle() Border {
...
+// on or off.
+func (s Style) implicitBorders() bool {
+	var (
+		borderStyle = s.getBorderStyle()
+		topSet      = s.isSet(borderTopKey)
+		rightSet    = s.isSet(borderRightKey)
+		bottomSet   = s.isSet(borderBottomKey)
+		leftSet     = s.isSet(borderLeftKey)
+	)
+	if borderStyle != noBorder && !(topSet || rightSet || bottomSet || leftSet) {
```
bashbunni, line 533: I think you can remove this `if` condition and just return
```suggestion
	return borderStyle != noBorder && !(topSet || rightSet || bottomSet || leftSet)
```

meowgorithm: Good catch!

get.go
```diff
@@ -519,6 +519,23 @@ func (s Style) getBorderStyle() Border {
...
+		borderStyle = s.getBorderStyle()
+		topSet      = s.isSet(borderTopKey)
+		rightSet    = s.isSet(borderRightKey)
+		bottomSet   = s.isSet(borderBottomKey)
+		leftSet     = s.isSet(borderLeftKey)
+	)
+	return borderStyle != noBorder && !(topSet || rightSet || bottomSet || leftSet)
+		return true
+	}
+	return false
```
aymanbagabas, line 536:
```suggestion
```

bashbunni: Thank you!

PR #417, opened 2024-10-29
(v2) feat!: make transform an interface type

aymanbagabas:
This change introduces a new `Transform` type with a canonical implementation `TransformFunc` that can be used to transform a string at render time. This change also updates the `Style` type to use the new `Transform` type instead of a function type.

We need this to support complex transforms and to be able to compare styles using the `==` operator.

### DIFF
--- a/set.go
+++ b/set.go
@@ -1,6 +1,8 @@
 package lipgloss
 
-import "image/color"
+import (
+	"image/color"
+)
 
 // Set a value on the underlying rules map.
 func (s *Style) set(key propKey, value interface{}) {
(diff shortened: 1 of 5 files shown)


set.go
```diff
@@ -1,6 +1,8 @@
 package lipgloss
 
-import "image/color"
+import (
+	"image/color"
+)
```
caarlos0, line 5:
```suggestion
import "image/color"
```

💅🏻

bashbunni (approved):

meowgorithm:
Functions can't be compared with `==` as (it's mathematically impossible). @aymanbagabas shall we close this one and reevaluate?

PR #437, opened 2024-11-15
fix: Color always returns TrueColor values

aymanbagabas:
This change removes magical ANSI color degradation from the Color function. Now, the function always returns TrueColor values. To use ANSI and ANSI256 colors, use lipgloss.Black, lipgloss.Red, lipgloss.BrightBlue, etc and lipgloss.ANSIColor values and type.

### DIFF
--- a/color.go
+++ b/color.go
@@ -51,9 +51,9 @@ func (n NoColor) RGBA() (r, g, b, a uint32) {
 	return 0x0, 0x0, 0x0, 0xFFFF //nolint:mnd
 }
 
-// Color specifies a color by hex or ANSI256 value. For example:
+// Color specifies a color by hex or int value. For example:
 //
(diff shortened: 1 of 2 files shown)


caarlos0 (approved):

meowgorithm:
So with this change, if you do something like the following…

```go
a := lipgloss.Color(212)
```

…is that essentially interpreted as 0x0000D4?

aymanbagabas:
> So with this change, if you do something like the following…
> 
> ```go
> a := lipgloss.Color(212)
> ```
> 
> …is that essentially interpreted as `0x0000D4`?

Correct

aymanbagabas:
Closing in favor of https://github.com/charmbracelet/lipgloss/pull/441

PR #451, opened 2024-12-02
fix!: style size to include borders

bashbunni:
included tests for these changes that fail on `master` but pass on this branch :)


examples/layout/main.go
```diff
@@ -126,9 +126,9 @@ var (
 	list = lipgloss.NewStyle().
 		Border(lipgloss.NormalBorder(), false, true, false, false).
 		BorderForeground(subtle).
-		MarginRight(2).
+		MarginRight(1).
 		Height(8).
-		Width(columnWidth + 1)
```
bashbunni, line 131: we don't need to do this anymore because the `+1` before was to compensate for the border on one side

examples/layout/main.go
```diff
@@ -282,7 +281,7 @@ func main() {
 				listItem("Pomelo"),
 			),
 		),
-		list.Width(columnWidth).Render(
```
bashbunni, line 285: all columns have the same width now

bashbunni:
@meowgorithm Should this change be part of v2 given it may break some layouts?

This change causes a lot of existing examples to be off by 1 or two depending on if the border is only on one side or not. Authors are currently working around by having `x +/- 1` or `x +/- 2` when working with bordered content. You can see this in our layout example.

That makes this change breaking despite it being the intended behaviour for `Width`...

bashbunni:
image
updated to use `v2-exp` as base branch

PR #453, opened 2024-12-09
fix(examples): use latest Color func

bashbunni:
Removed octal format from examples. It's still mentioned outside of examples including in tests.

bubble tea examples are broken from changes in ansi (e.g. `examples/color/bubbletea/`)


meowgorithm:
This is super. Thanks, Bash!

PR #454, opened 2024-12-10
feat(tree): hide children

bashbunni:
I updated the godoc of tree to match capitalization used for leaf while I was at it


caarlos0 (approved):

aymanbagabas (approved):

bashbunni:
Another thing to improve this PR: have an example hiding both ways with comments on why you would use Hide vs SetHidden. It already says in their go doc, but could be handy to show how they differ in usage.

bashbunni:
SetValue for tree does the same thing as Root. Should I mention any difference in the docs?

tree/tree.go
```diff
@@ -121,6 +142,11 @@ func (t *Tree) Value() string {
 	return t.value
 }
 
+// SetValue sets the value of a Tree node.
+func (t *Tree) SetValue(value any) {
+	t = t.Root(value)
```
caarlos0: probably need to be something like `*t = t.Root(value)`

caarlos0 (approved):

bashbunni:
After looking at this again and putting an example together, I think we may want a `SetChildren` method for `Node`. If I just set the value on a leaf, it overwrites the existing value with the new tree (which makes sense). So we're forced to get the value of the node we're modifying before modifying it. See below

```go
	enumeratorStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("63")).MarginRight(1)
	rootStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("35"))
	itemStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("212"))

	t := tree.
		Root("⁜ Makeup").
		Child(
			"Glossier",
			"Fenty Beauty",
			tree.New().Child(
				"Gloss Bomb Universal Lip Luminizer",
				"Hot Cheeks Velour Blushlighter",
			),
			"Nyx",
			"Mac",
			"Milk",
		).
		Enumerator(tree.RoundedEnumerator).
		EnumeratorStyle(enumeratorStyle).
		RootStyle(rootStyle).
		ItemStyle(itemStyle)
	glossier := t.Children().At(0)
	glossier.SetValue(tree.Root(glossier.Value()).Child(tree.Root("Apparel").Child("Pink Hoodie", "Baseball Cap")))
	fmt.Println(t.String())
```

image

---

Another thing I noticed here is that when you add a nested tree after the original tree was created, it doesn't inherit the styles. Check out the difference between Glossier and Fenty Beauty above. Will work on that in a separate PR using this as the base branch

bashbunni:
I simplified the usage and it feels much better now. You can update a `Tree`'s children pretty easily. The changes include:
- overwrite all existing children
- insert a child at a specific index
- replace a child at a specific index

and these work with `Leaf` types too by converting the `Leaf` to a `Tree` type.

Here's what it looks like:
```go
        enumeratorStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("63")).MarginRight(1)
	rootStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("35"))
	itemStyle := lipgloss.NewStyle().Foreground(lipgloss.Color("212"))

	t := tree.
		Root("⁜ Makeup").
		Child(
			"Glossier",
			"Fenty Beauty",
			tree.New().Child(
				"Gloss Bomb Universal Lip Luminizer",
				"Hot Cheeks Velour Blushlighter",
			),
			"Nyx",
			"Mac",
			"Milk",
		).
		Enumerator(tree.RoundedEnumerator).
		EnumeratorStyle(enumeratorStyle).
		RootStyle(rootStyle).
		ItemStyle(itemStyle)
	// Add a Tree as a Child of "Glossier"
	t.Replace(0, t.Children().At(0).Child(
		tree.Root("Apparel").Child("Pink Hoodie", "Baseball Cap"),
	))

	// Add a Leaf as a Child of "Glossier"
	t.Children().At(0).Child("Makeup")
	fmt.Println(ansi.Strip(t.String()))
```

Note: you have to call `t.Replace` on the parent if you're trying to modify its child Leaf. Because it's changing type, we need to reassign the value for the parent (See "Add child to Glossier" above compared to following `Child` operation that's now working on a `Tree`) - Will add this note to the godoc for `Leaf` operations

bashbunni:
I'll split this PR into two... The other `Leaf/Tree/NodeChildren` modifications are probably out of scope here

bashbunni:
Simplified and split this into separate PRs https://github.com/charmbracelet/lipgloss/pull/461 & https://github.com/charmbracelet/lipgloss/pull/460

PR #458, opened 2025-01-07
feat: style ranges

caarlos0:
Extracted from https://github.com/charmbracelet/gum/pull/789 , this allows to style ranges of a given string without breaking its current styles.

The resulting ansi sequences aren't that beautiful (as there might be many styles+reset with nothing in them), but it works. We can optimize this later I think.

Also: maybe better waiting for lipgloss v2 and do it there instead. This does not handle renderers for example.

### DIFF
--- a/ranges.go
+++ b/ranges.go
@@ -8,6 +8,7 @@ import (
 
 // StyleRanges allows to, given a string, style ranges of it differently.
 // The function will take into account existing styles.
+// Ranges should not overlap.
 func StyleRanges(s string, ranges []Range) string {
 	if len(ranges) == 0 {
 		return s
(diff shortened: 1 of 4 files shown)


ranges.go
```diff
@@ -0,0 +1,52 @@
...
+	for _, rng := range ranges {
+		// Add the text before this match
+		if rng.Start > lastIdx {
+			buf.WriteString(ansi.Cut(s, lastIdx, rng.Start))
+		}
+		if l := len(stripped); rng.End >= l {
+			rng.End = l - 1
+		}
+		// Add the matched range with its highlight
+		buf.WriteString(rng.Style.Render(stripped[rng.Start : rng.End+1]))
```
aymanbagabas, line 31: My concern here is dealing with multibyte runes and multi-rune characters. Not sure what's the behavior with such cases.

caarlos0: right, yeah, probably is not correct indeed

ranges.go
```diff
@@ -6,6 +6,20 @@ import (
 	"github.com/charmbracelet/x/ansi"
 )
 
+// StyleRanges allows to, given a string, style a range of it differently.
+// The function will take into account existing styles.
+// See [StyleRanges] to style multipe ranges in the same string.
```
meowgorithm, line 11: Ugh why do I need to comment before making a code suggestion?

meowgorithm:
```suggestion
// See [StyleRanges] to style multiple ranges in the same string.
```

bashbunni (approved):

ranges.go
```diff
@@ -0,0 +1,62 @@
...
+import (
+	"strings"
+
+	"github.com/charmbracelet/x/ansi"
+)
+
+// StyleRanges allows to, given a string, style a range of it differently.
+// The function will take into account existing styles.
+// See [StyleRanges] to style multipe ranges in the same string.
+func StyleRange(s string, start, end int, style Style) string {
```
aymanbagabas, line 12: Instead, maybe we could do `StyleRanges(s, NewRange(start, end, style))` where `func StyleRanges(s string, ranges ...Range) string` 🤔

caarlos0: yeah sounds good to me as well :)

meowgorithm: Yeah, great call @aymanbagabas. Part of me thinks it should be singular, though, just so users don't think it must take multiple values 🤔. But maybe I'm overthinking it.

caarlos0: fwiw I'm more inclined into using plural, as it indicates one or more, whereas singular indicates only one...

aymanbagabas (approved):

PR #460, opened 2025-01-10
feat(tree): hide children

bashbunni:
See examples for usage on these changes. Those examples are testable, so see their output in comment to see how these API changes work.


bashbunni:
@dlvhdr based on your work on the tree bubble, it would be great to get your thoughts on this PR and #461

aymanbagabas (approved):

PR #461, opened 2025-01-10
feat(tree): modify children of existing `Tree`

bashbunni:
This allows users to modify existing `Trees`. In the current version of this table, there is no way to change existing children outside of adding a new child with `Tree.Child()`.

This also handles operations on `Leaf` types to turn them into `Trees`.

See examples for usage on these changes. Those examples are testable, so see their output in comment to see how these API changes work.


tree/tree.go
```diff
@@ -118,7 +151,64 @@ func (t *Tree) String() string {
...
+		case *Tree:
+			child, _ := child.(*Tree)
+			nodes = append(nodes, child)
+		case Children:
+			for i := 0; i < item.Length(); i++ {
+				nodes = append(nodes, item.At(i))
+			}
+		case Node:
+			nodes = append(nodes, item)
+		case fmt.Stringer:
```
aymanbagabas, line 188: Might be useful to support `encoding.TextMarshaler` as well

bashbunni: We don't have this in other type checks in `tree`, maybe something we can add down the line

bashbunni:
Need to look at the feedback from https://github.com/charmbracelet/lipgloss/discussions/452 more closely

PR #466, opened 2025-01-24
wip: Column wrapping

bashbunni:
This PR fixes improper content truncation in lipgloss table. See https://github.com/charmbracelet/glamour/issues/344

You can run `go test -run TestContentWrapping -v` to see the current output

before:
image

current:
image

TODO:
- [ ] wrap headers if they're long?
- [ ] wrap content if no height is set?
- [ ] more tests for edge cases (wip)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/charmbracelet/x/ansi v0.6.1-0.20250107110353-48b574af22a5
 	github.com/charmbracelet/x/exp/golden v0.0.0-20240806155701-69247e0abc2a
 	github.com/muesli/termenv v0.15.2
+	github.com/olekukonko/tablewriter v0.0.5
 	github.com/rivo/uniseg v0.4.7
 )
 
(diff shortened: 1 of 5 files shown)


bashbunni:
Still need to iron out some details with the team first, but added some tests that show behaviour we want to modify. They're not failing, just logging outputs because I don't know what the desired output is yet for those cases.

bashbunni:
Based on behaviour of html tables, it wraps by default, but you can force it to truncate content too. Html tables will wrap and cut content if a height + overflow is set. In our case, I'll truncate content to fit in a given height if one is given.

If we're wrapping content, we'll need to keep track of the tallest cell in each row as that will be the height of that row.

testinfected:
Thanks for working on this. Any way I can help with this?

bashbunni:
Hey @testinfected nothing at the moment. This is a bit of a can of worms, so it might take some more time. In the meantime, are you able to use the old version of Glamour if the table is working properly for you? We are expediting this one, but there's a good chunk of code that needs a re-write to fix this properly.

Most of those changes are now happening on a new branch, so I'm going to delete this one along with the PR, but there will be a new PR up asap to replace this one 🙏

testinfected:
Sure we can live with 0.7 for the time being. Let's do the thing right!

bashbunni:
@testinfected thanks for understanding :) we will do our best!

PR #471, opened 2025-02-11
feat: initial compositing implementation

aymanbagabas:
This adds a new `Canvas` type that can be used to compose multiple layers together. Each layer can have its own content, position, and z-index. The `Canvas` type can render the layers to a string.

https://github.com/user-attachments/assets/392dc82f-7658-4797-998b-f45f5d9eef22

Screenshot 2025-02-11 at 4 49 04 PM


canvas.go
```diff
@@ -0,0 +1,279 @@
+package lipgloss
```
raphamorim, line 1: The example of the PR is in the source code?

raphamorim: nvm just saw your second commit 🙏

raphamorim (approved):

andreynering (approved):
This looks so cool! 👏

meowgorithm:
I've pulled the Bubble Tea example as it records view state to the model, which we should be handling natively in Bubble Tea. Otherwise, from a Lip Gloss perspective, I believe this is good to go.

meowgorithm (approved):

Issue #431, opened 2024-11-08
Breaking Change in Header Styling from Index 0 to Index -1 in Tag 0.13.0 -> 0.13.1


joaogmguimaraes:
From tag `0.13.0` to tag `0.13.1`, there was a breaking change where the header styling in the table changed from index 0 to index -1 (a great improvement, by the way). However, this affected several scripts after updating the library.

I suggest we use the [Semver](https://semver.org/) standard for new tags `1.0.1`, ..., `2.1.0` to avoid surprises, especially for developers who use `go get -u=patch ./...`.

joaogmguimaraes:
I almost forgot, the README.md and the code comments became outdated after this change.

meowgorithm:
Thanks, @joaogmguimaraes, and apologies about the change in API. We'll put this on our to-do list, but feel free to open a PR with the fixes if you'd like, too.

PR #475, opened 2025-02-13
docs: fix table example on readme

andreynering:
1. The header index is actually `-1` and not `0`. Checking for the `table.HeaderRow` const which is set to `-1`.
1. Adding the style definitions so the example is actually complete and can be copy & pasted and ran more easily.

### DIFF
--- a/README.md
+++ b/README.md
@@ -425,17 +425,28 @@ rows := [][]string{
 Use the table package to style and render the table.
 
 ```go
+var (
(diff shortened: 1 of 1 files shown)


bashbunni:
Fixes: https://github.com/charmbracelet/lipgloss/issues/431

We should also double check the godoc examples work

aymanbagabas (approved):

PR #476, opened 2025-02-14
test(tree): use golden file to preserve ANSI sequences

bashbunni:
This might be a "fix" rather than "test" conventional commit, but this fixes a test that is broken on master. Not sure if/how this ever passed since it was giving a styled output despite the `want` value not being styled. This change uses a golden file to track the style changes (resulting ANSI sequences) in the test output.

Didn't work on `v1.0.0` either 🤷


bashbunni:
Oh right, it's because the terminal used for GitHub actions doesn't have colour capability, so the test was passing with GitHub actions and failing locally... Now this fails CI 😂 We can either add `lipgloss.SetColorProfile(termenv.TrueColor)
` at the beginning of this test to force coloured output in GH actions or not use a golden file and compare `want` to `ansi.Strip(table.String())` and ignore the colours in the test

aymanbagabas (approved):

meowgorithm (approved):

PR #479, opened 2025-02-24
feat(table): improve sizing and behavior: wrap by default, overflow optionally

andreynering:
Hey everybody,

The whole work took me multiple days to get right, with a lot of testing and back-and-fourth.

I think it is very close to the ideal now. We may want to do small adjustments, and in this regard I ask for your opinion.

Highlights:

- The whole resizing logic was moved into a separate file, using multiple functions. Before, it was all done inside the  `Table.String()` function.
- Wrapping will now happen by default.
- If wrapping is disabled with `t.Wrap(false)`, the behavior of adding a trailing `…` is now fixed. Before, it was just cutting the content most of the time.

---

> [!NOTE]
> There are many issues related to table rendering. I think this PR might fix all of them. I vote to close all these issues to clean the list up, and if any user still face any minor issue, they can still comment or open new issues to let us know.

- Closes https://github.com/charmbracelet/lipgloss/issues/459
- Closes https://github.com/charmbracelet/glamour/issues/344
- Closes https://github.com/charmbracelet/lipgloss/issues/465
- Closes https://github.com/charmbracelet/lipgloss/issues/468
- Closes https://github.com/charmbracelet/lipgloss/issues/406
- Closes https://github.com/charmbracelet/lipgloss/issues/400
- Closes https://github.com/charmbracelet/lipgloss/issues/381
- Closes https://github.com/charmbracelet/lipgloss/issues/355
- Closes https://github.com/charmbracelet/lipgloss/issues/261
- Previous work: https://github.com/charmbracelet/lipgloss/pull/466
- Previous work: https://github.com/charmbracelet/lipgloss/compare/master...table-auto-wrap

---

Screenshots

<details><summary>Before</summary>

Screenshot 2025-02-24 at 17 37 36

</details> 

<details><summary>After</summary>

Screenshot 2025-02-24 at 17 33 48

Screenshot 2025-02-24 at 17 35 06

</details>

PR #481, opened 2025-02-26
chore: add .editorconfig

aymanbagabas:
This adds an .editorconfig file to the repository to help maintain consistent formatting across the codebase. Particularly, it sets the final newline behavior for golden files. This is based on https://github.com/charmbracelet/lipgloss/pull/479

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,18 @@
+# https://editorconfig.org/
+
+root = true
+
+[*]
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 2
+
+[*.go]
+indent_style = tab
+indent_size = 8
+
+[*.golden]
+insert_final_newline = false
+trim_trailing_whitespace = false
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+*.golden linguist-generated=true -text
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,4 +1,0 @@
-{
-  "files.trimTrailingWhitespace": false,
-  "files.insertFinalNewline": false
-}


.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
+charset = utf-8
+
+[table/table_test.go, *.golden]
```
raphamorim, line 7: `*.golden` ?

.editorconfig
```diff
@@ -0,0 +1,9 @@
+root = true
+
+[*]
+end_of_line = lf
```
raphamorim, line 4: suggestions to add:

```suggestion
end_of_line = lf
indent_style = space
indent_size = 2

[*.md]
indent_size = 4
trim_trailing_whitespace = false
```

.editorconfig
raphamorim, line 7: why not `*.go`?

raphamorim: if make sense prob will be good to include `go.mod,go.sum,*.go`

andreynering (approved):

.editorconfig
andreynering, line 7: Ideally, we should refactor all table tests to use `.golden` files. Then, we won't need a special rule for `table/table_test.go` anymore.

aymanbagabas: Golden files are used for testing

raphamorim: alright, it's used for testing

.editorconfig
```diff
@@ -0,0 +1,9 @@
...
+charset = utf-8
+
+[table/table_test.go, *.golden]
+trim_trailing_whitespace = false
+insert_final_newline = false
```
andreynering, line 9: Reverting my accept in favor of some suggestions, actually:

* In VSCode, `[table/table_test.go, *.golden]` does not work. It doesn't work for `table_test.go` and break the rule for `*.golden`. `[*.golden]` is what works.
  * I tried to duplicate the rule for `[table/table_test.go]` but for some reason that does not work either.
* Making everything 2-space by default and only Go files `8-tab`.
* I do not recommend `end_of_line = lf` as Git for Windows handles that automatically. There is a Git option to control how files should be checked out and commmited.

```suggestion
# https://editorconfig.org/

root = true

[*]
charset = utf-8
insert_final_newline = true
trim_trailing_whitespace = true
indent_style = space
indent_size = 2

[*.go]
indent_style = tab
indent_size = 8

[*.golden]
insert_final_newline = false
trim_trailing_whitespace = false
```

andreynering (changes requested):

andreynering (approved):

.editorconfig
andreynering: NOTE: Pushed the fixes myself as we agreed on Discord.

PR #483, opened 2025-03-04
test(table): add overflow test + use golden files

bashbunni:
Tidied up some of the new wrapping tests to have the test cases follow nomenclature in other table tests in this package. Also made the tests use golden files as they were previously just logging the outputs, so would always pass. Added a test for manually setting widths on columns. These tests are failing as I don't think they are overriding the width values properly.

That said, we can probably discuss what column width ranges we might want to support? tbd


andreynering:
Hey @bashbunni,

Thank you for working on it! There was indeed a bug where we were ignoring fixed column widths.

Cascading PRs can be a bit messy on GitHub. Unless we avoid force pushes and use merge (instead of squash or rebase) on the base branch, we can get a messy diff and need to reapply the commits on the other PRs.

I just cherry-picked your commits from here to #479 with fixes and pushed them. All tests are now passing. You might want to take a look there. If you think it's better to keep a separate PR, we can still delete those commits from there and cherry-pick into a new PR, but IMO these commits belongs to #479.

bashbunni:
Hey @andreynering thank you! I'll close this one. I just wanted to give you some way to review + explain my changes before changing things in your branch 

Thanks for taking this on by the way :)

PR #489, opened 2025-03-12
fix: use non-breaking space for padding

aymanbagabas:
Padding should use a non-breaking space so that the padding is preserved when the string is being copied and pasted, and wrapped to the next line.

### DIFF
--- a/style.go
+++ b/style.go
@@ -524,7 +524,9 @@ func pad(str string, n int, style *ansi.Style) string {
 		return str
 	}
 
-	sp := strings.Repeat(" ", abs(n))
+	// We use a non-breaking space to pad so that the padding is
+	// preserved when the string is copied and pasted.
+	sp := strings.Repeat("\u00a0", abs(n))
 	if style != nil {
 		sp = style.Styled(sp)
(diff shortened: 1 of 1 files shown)


style.go
```diff
@@ -524,7 +524,9 @@ func pad(str string, n int, style *ansi.Style) string {
 		return str
 	}
 
-	sp := strings.Repeat(" ", abs(n))
+	// We use a non-breaking space to pad so that the padding is
```
raphamorim, line 527: this make sense but we are sure we want this for all the cases?

no-break space is used to introduce a space where you don't want line breaks, like where it'll look strange or easily be misunderstood if a line break.

aymanbagabas: That's why we want them here in padding.

raphamorim: ah right, sorry had to ask just to double check

raphamorim (approved):

meowgorithm (approved):
This is a solid one. Let's get it into the release today.

andreynering:
CI failing. Tests need to be updated.

Issue #488, opened 2025-03-10
Colored and styled underlines


aymanbagabas:
**Is your feature request related to a problem? Please describe.**
Many terminals and programs such as Kitty, Ghostty, iTerm2, and Neovim support styled underlines. This is described in details here https://sw.kovidgoyal.net/kitty/underlines/

**Describe the solution you'd like**
Lip Gloss should be able to change underline style and color.

**Describe alternatives you've considered**
Using `x/ansi` directly.

**Additional context**
https://sw.kovidgoyal.net/kitty/underlines/

meowgorithm:
This looks good, but let's target v2.x. This should be implemented with `/x/ansi`, correct?

aymanbagabas:
> This looks good, but let's target v2.x. This should be implemented with `/x/ansi`, correct?

Correct, `x/ansi` already supports this, we just need to figure out how the API would work with `style.Underline(bool)`. Do we want `style.UnderlineStyle(...)` or since this is targeting v2, do we want to repurpose `style.Underline(UnderlineStyle, color.Color)`? How would we support underline colors? `style.UnderlineColor(...)` is not consistent with `style.Foreground(...)`

Things to keep in mind while thinking about this 🙂

PR #491, opened 2025-03-12
feat: add underline style and color

aymanbagabas:
This adds support for setting the underline style and color. The underline style can be set to single, double, curly, dotted, or dashed. The color of the underline can be set to any color.

See https://sw.kovidgoyal.net/kitty/underlines/

Fixes: https://github.com/charmbracelet/lipgloss/issues/488


raphamorim (approved):

aymanbagabas:
> Also a question: how does this behave in unsupported terminals? Will it fall back to having a normal underline? If not, can we make it fall back accordingly?

It depends on the terminal and how they interpret the sequence. Apple Terminal for example, behaves incorrectly and doesn't support neither underline styles nor underline colors.

aymanbagabas:
> @aymanbagabas in that case what happens? Does it underline at all?

No, it doesn't underline and instead writes the text with a background color.

PR #492, opened 2025-03-14
feat: Vertical Cell Merging In Table Columns

loubenl:
# Changes
Merge adjacent vertical cells in the columns specified by a new []int field in the Table struct called borderMergeColumns. 

Borders in between adjacent vertical cells containing identical data in these columns would be removed, where only the topmost cell's contents would remain with all other cell contents being hidden.

# Example
Merge the first and last column
Screenshot 2025-03-14 at 2 29 20 PM

Screenshot 2025-03-14 at 2 28 47 PM

# Details
This functionality would only work if both borderColumn and borderRow are set to true. For example, if a user wanted to merge vertical cells in columns 0, 2, 3, when creating the table, the user would need to add table.New().borderColumn(true).borderRow(true).BorderMergeColumns(0,2,3)

All other functionality would remain the same : borderColumn(true).borderRow(true) would just draw all borders, borderColumn(true).borderRow(false) would only draw the column borders... etc

# Testing 
All unit tests are in `lipgloss/table/table_borders_test.go`


meowgorithm:
Hi! Thanks for the contribution. This is an impressive implementation, however we're unlikely to merge it in the short term because of the upcoming [v2](https://github.com/charmbracelet/lipgloss/releases/tag/v2.0.0-beta.1) release (maintained on the `v2-exp` branch) and because of the impact on the codebase.

Issue #403, opened 2024-10-18
use golden files for all tests


bashbunni:
- [ ] replace desired results in tests with golden files
- [ ] have a script to clean up old golden files? When a test or subtest is renamed, it creates a new golden file, but won't clean up the old files.

PR #503, opened 2025-03-20
test: migrate all table tests to golden files

andreynering:
- Closes https://github.com/charmbracelet/lipgloss/issues/403

This is not only more standard, but also prevent issues with editors stripping meaningful trailing white-space from `table_test.rb`, which make some tests fail.


aymanbagabas (approved):

bashbunni:
Thanks so much for this Andrey :)

PR #508, opened 2025-03-31
fix(table): fix panic when style func is `nil`

### LINKED ISSUE #468: bug(table): `computeHeight` doesn't honor all rows when hasHeaders is false
Hello,

The [`computeHeight`](https://github.com/charmbracelet/lipgloss/blame/ec86e885226610493976264f4c840401ca1d96c7/table/table.go#L406) function doesn't seem to honor the number of rows returned by the `Data.Rows()` interface method.

I have implemented the [`table.Data`](https://pkg.go.dev/github.com/charmbracelet/lipgloss/v2/table#Data) interface to return the number of rows my table bubble contains using the `Rows()` method.

I have noticed that when I "hide" the headers by using `table.Headers()`, then the `computeHeight` function returns a value that doesn't represent the correct number of rows.

I believe this is due to the fact that the function substracts `1` from the returned height, which I cannot understand why: `return sum(t.heights) - 1 + btoi(hasHeaders) +`.

When `hasHeaders` is true, it essentially "cancels out" the previous substraction, and thus the number of rows that will be displayed is correct, since the returned value is used to [define the MaxHeight](https://github.com/charmbracelet/lipgloss/blob/ec86e885226610493976264f4c840401ca1d96c7/table/table.go#L391).

I have confirmed that if I "artificially increase" the number of rows returned by my `Data.Rows()` method implem, then the height of the table is correct and all rows are displayed, however this implies that the `x` index passed to the `Data.At` method is wrong and leads to out of bounds panics.

andreynering:
- Fixes #468

### DIFF
--- a/table/resizing.go
+++ b/table/resizing.go
@@ -59,6 +59,11 @@ func (t *Table) resize() {
 		allRows = rows
 	}
 
+	styleFunc := t.styleFunc
+	if t.styleFunc == nil {
+		styleFunc = DefaultStyles
+	}
+
 	r.rowHeights = r.defaultRowHeights()
 
 	for i, row := range allRows {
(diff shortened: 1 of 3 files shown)


raphamorim (approved):
Considering is ok to use the default then 👌 lgtm

caarlos0 (approved):

PR #512, opened 2025-04-04
refactor: v2 rename offset

bashbunni:
This matches what we use in bubbles and allows us to potentially support an `xOffset` in the future.


table/table.go
```diff
@@ -210,9 +210,9 @@ func (t *Table) Height(h int) *Table {
...
 }
 
-// Offset sets the table rendering offset.
-func (t *Table) Offset(o int) *Table {
-	t.offset = o
+// YOffset sets the table rendering offset.
+func (t *Table) YOffset(o int) *Table {
+	t.yOffset = o
 	return t
 }
```
andreynering, line 217: In theory you could keep the original `Offset` function and add a comment marking it as deprecated:

https://go.dev/wiki/Deprecated

No strong opinion, just commenting about a possibility.

andreynering: Sure, it's fine to remove it IMO

aymanbagabas (approved):

andreynering (approved):

Issue #524, opened 2025-05-11
Function description mismatch - table.New


dunric:
`func lipgloss.table.New() *lipgloss.table.Table`
> New returns a new Table that can be modified through different attributes.
>
> By default, a table has no border, no styling, and no rows.

while it actually *does* set (round) border around:

https://github.com/charmbracelet/lipgloss/blame/42d933d1dad1715059e427db8c31ab4f08ec4d26/table/table.go#L74-L87

andreynering:
Hi @dunric,

We changed to `NormalBorder` on v2 and update the docs on #525.

PR #525, opened 2025-05-12
chore(table): set normal borders by default on v2 + fix docs

andreynering:
- Closes https://github.com/charmbracelet/lipgloss/issues/524

Consider this PR a proposal for discussion. I have no strong opinions.

Since the beginning, the table has rounded borders by default, but the docs says it has no borders. To me, it doesn't make much sense to have no borders by default, so I propose to either set "normal borders" by default starting on v2 and adjust the docs, or to just mention "rounded corners" as the default on the docs.

Normal and rounded border look particularly similar on some fonts, like on JetBrain's Mono. In theory "normal" is saner as a default, but rounded looks prettier.


meowgorithm:
Dude, _great_ catch. I agree with this change.

bashbunni:
I'm good with whatever, but I do think it makes sense to have the default be somewhat "charm branded". That is a big perk of our libraries in general: backend devs not having to do much to make it look pretty out of the box. I think it should be rounded by default purely because it looks nicer and most UIs these days round borders by default (GitHub, YouTube, Figma, Apple, etc).

My 2c

PR #529, opened 2025-05-29
fix(table): fix resizer ignoring borders when deciding whether to shrink

MartinodF:
### Describe your changes

`resizer.maxTotal()` is compared to `r.tableWidth` within `resizer.optimizedWidths()` to decide whether to shrink or expand the column sizes according to the available space.

Unfortunately, `maxTotal()` only takes into account the column sizes (fixed width if specified, max character count + padding otherwise), ignoring the space taken up by borders.

This leads to the table resizing breaking as follows: if a table has 1 column with a max character count of 10:
- Setting a table width < 8 works as expected
- Setting a table width >= 8 && < 10 leads to `expandTableWidth()` being called even if the table needed to be shrunk, causing the contents to be cut off
- Setting a table width > 10 works as expected

Change `maxTotal()` to include `r.totalHorizontalBorder()` so that borders are taken into account and the table is shrunk up until the width where borders + column widths fit.

### Checklist before requesting a review

- [x] I have read `CONTRIBUTING.md`

### DIFF
--- a/table/resizing.go
+++ b/table/resizing.go
@@ -360,7 +360,8 @@ func (r *resizer) maxCharCount() int {
 }
 
 // maxTotal returns the maximum total width.
-func (r *resizer) maxTotal() (maxTotal int) {
+func (r *resizer) maxTotal() int {
+	maxTotal := r.totalHorizontalBorder()
 	for j, column := range r.columns {
 		if column.fixedWidth > 0 {
 			maxTotal += column.fixedWidth


andreynering (commented):
Hi @MartinodF,

Yes, that should be fixed on #526. I also want to avoid having conflicts with that branch.

It'd be nice if you tested that branch and let us know if it works fine for you.

Also, note for future PRs that you should usually target `v2-exp` instead of `master` (until v2 is released).

MartinodF:
@andreynering makes perfect sense. I've been delaying the switch from bubbles/table to lipgloss/table ever since it was released because of this bug, today I had a bit of free time and decided to contribute a fix without noticing your PR.

I can confirm #526 solves this, I'll wait for a release then. Thanks!

Issue #534, opened 2025-06-03
[v2] padding gets rendered as Â


tom-wegener:
**Describe the bug**

lipgloss Padding gets rendered as Â.

**Setup**
Please complete the following information along with version numbers, if applicable.
 - OS: debian 11
 - Shell: dash probably
 - Terminal Emulator: tty
 - started vm with debian os and the tool only

**To Reproduce**

**Source Code**
Please include source code if needed to reproduce the behavior. 

**Expected behavior**
padding should be rendered as whitespace.

**Screenshots**

![Image](https://github.com/user-attachments/assets/d5c60586-f6b4-41a8-908a-817497104591)

**Additional context**

tom-wegener:
the output of locale is the following:
```
root@foo-bar:~# locale
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_ NUMERIC="POSIX"
LCLTIME="POSIX"
LC_COLLATE="POSIX"
LC_MONE TARY="POSIX
LC_MESSAGES="POSIX
LC PAPER="'POSIX"
LO_NAME="POSIX**
LC-ADDRESS="POSIX"
LC_TELEPHONE= "POSIX"
LCIMEASUREMENT="POSIX"
L_IDENTIFICATION="POSIX"
LC_ALL=
root@foo-bar:~#
```

prithvijj:
Ah alrighty, thought it might be something related to setting `export LANG=en_US.UTF-8` to help with things,
But yeah seems like there's a PR created to fix it?

tom-wegener:
> Ah alrighty, thought it might be something related to setting `export LANG=en_US.UTF-8` to help with things, But yeah seems like there's a PR created to fix it?

Thank you, I'll test it in the coming days/weeks and will close this ticket then.

PR #545, opened 2025-07-04
fix: revert back to regular space

caarlos0:
using ` ` (`\u00a0` - non breaking space) instead of a regular space seems to be causing all sorts of problems:

- https://github.com/charmbracelet/fang/issues/57
- https://github.com/charmbracelet/lipgloss/issues/534

Maybe it's best to revert back to using regular spaces instead? Or make it opt-in? 

refs https://github.com/charmbracelet/lipgloss/pull/489

### DIFF
--- a/set.go
+++ b/set.go
@@ -324,7 +324,7 @@ func (s Style) PaddingBottom(i int) Style {
 	return s
 }
 
-// PaddingBottom adds padding to the bottom of the block.
+// PaddingChar sets the character used for padding.
 func (s Style) PaddingChar(c rune) Style {
 	s.set(paddingCharKey, c)
 	return s


style.go
```diff
@@ -10,7 +10,7 @@ import (
 )
 
 const (
-	nbsp            = '\u00A0'
+	nbsp            = ' '
```
aymanbagabas, line 13:
```suggestion
```

style.go
```diff
@@ -395,9 +395,6 @@ func (s Style) Render(strs ...string) string {
 			str = padLeft(str, leftPadding, st, nbsp)
```
aymanbagabas, line 395:
```suggestion
			str = padLeft(str, leftPadding, st, ' ')
```

PR #546, opened 2025-07-07
feat: add padding and margin character support

aymanbagabas:
This adds support for setting a padding and margin character in the Style struct. This allows users to specify a character that will be used for padding and margin and defaults to whitespace if not set.


caarlos0:
superseedes https://github.com/charmbracelet/lipgloss/pull/545

PR #550, opened 2025-07-18
feat: colors sub-package: color blending & other low-level color utilities

lrstanley:
## Describe your changes

- Initial implementation of sub-package `colors`, which provides helpers for mutating colors, as well as adding color blending/gradient functionality, including:
  - `BlendLinear1D` - 1-dimensional linear blending between 2+ color stops across `N` number of steps.
    - Has examples.
  - `BlendLinear2D` - 2-dimensional linear blending between 2+ color stops across `N` number of steps, with an angle to control direction (e.g. diagonal).
    - Has examples.
  - `Lighten` - increase the brightness of a color.
    - Has examples (lighten/darken in 1).
  - `Darken` - decrease the brightness of a color.
    - Has examples (lighten/darken in 1).
  - `Alpha` - simple helper for adjusting the alpha value of a color.
  - `Complementary` - get the complementary color of a given color.

If this is merged, this will also render the following no longer necessary:

- `github.com/charmbracelet/x/exp/color`
- `github.com/charmbracelet/x/exp/charmtone#Blend`

## Screenshots & Examples

<details>
<summary>All screenshots & videos</summary>

Video is a bit compressed, but:

https://github.com/user-attachments/assets/3f6d7263-16f0-44b1-8ef8-32914ab8d4f2

image

image

image

</details>

## Related issue/discussion:

- ref: https://github.com/charmbracelet/lipgloss/pull/154#issuecomment-2738045085
- ref: https://github.com/charmbracelet/lipgloss/discussions/50

## Checklist before requesting a review

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)

## If this is a feature

- [x] I have created a discussion (n/a: discussed with @meowgorithm)
  - Discussed in Discord with Christian. tl;dr: I don't think the charm team originally wanted blend functionality grouped in with lipgloss (re: [154](https://github.com/charmbracelet/lipgloss/pull/154#issuecomment-2738045085) & [50](https://github.com/charmbracelet/lipgloss/discussions/50), however, the usage of blending continues to grow in the bubbletea/lipgloss ecosystem, and there is now interest in including this in lipgloss directly for simplicity.
- [x] A project maintainer has approved this feature request.

### HCL vs LAB

The `Blend*` functionality uses `BlendLab` rather than `BlendHcl` (what `github.com/charmbracelet/x/exp/color` and crush use). For both 1D and 2D implementations, when using HCL, I noticed that despite proper input into the `BlendHcl` method, for some colors, it produced output which did not gracefully blend multiple colors (one color standing out drastically from others, see below screenshot and example). In addition, it would often cut off a good portion of the start and end colors as part of gradient generation. Using LAB, it's much more consistent and doesn't seem to suffer from the issues with the `BlendHcl` implementation.

<details>
<summary>Color blend bug with BlendHcl</summary>

image
</details>

To replicate this issue, switch `BlendLab` with `BlendHcl`, and use this repro:

<details>
<summary>BlendHcl reproduction code</summary>

```go
package main

import (
	"image/color"
	"strings"

	"github.com/charmbracelet/lipgloss/v2"
	"github.com/charmbracelet/lipgloss/v2/colors"
)

func main() {
	height := 15
	width := 65
	angle := 0

	blendedColors := colors.BlendLinear2D(
		width,
		height,
		angle,
		color.RGBA{R: 0, G: 119, B: 182, A: 255},
		color.RGBA{R: 72, G: 202, B: 228, A: 255},
		color.RGBA{R: 173, G: 232, B: 244, A: 255},
	)

	var s strings.Builder
	for y := range height {
		for x := range width {
			index := y*width + x
			blockStyle := lipgloss.NewStyle().Foreground(blendedColors[index])
			s.WriteString(blockStyle.Render("█"))
		}
		if y < height-1 {
			s.WriteString("\n")
		}
	}

	lipgloss.Println(s.String())
}
```

</details>

### Usage of 1D return array for `BlendLinear2D`

With the current implementation of `BlendLinear2D`, I return a 1D array in row-major order. This is primarily due to performance; however, it does feel a little more annoying to use:

<details>
<summary>1D row-major order loop pseudo-code</summary>

```go
gradient := colors.BlendLinear2D(width, height, 180, color1, color2, color3, ...)
gradientContent := strings.Builder{}
for y := range height {
	for x := range width {
		index := y*width + x
		gradientContent.WriteString(
			lipgloss.NewStyle().
				Background(gradient[index]).
				Render(" "),
		)
	}
	if y < height-1 { // End of row.
		gradientContent.WriteString("\n")
	}
}
```

</details>

I benchmarked returning a 2D array, which would be simpler to iterate over, however, results in about **49% more** allocations and **10% more** memory usage for a 100x100 test case. 100x100 is of course probably a much higher amount than someone would actually use, however, even for the examples I've provided, gradient generation (although I think as simple as it can be as far as allocations), still does use a good amount of memory _if you're not caching the gradient in your model -- e.g. dynamic usecases_.

<details>
<summary>2D row-major order loop pseudo-code</summary>

```go
gradient := colors.BlendLinear2D(width, height, 180, color1, color2, color3, ...)
gradientContent := strings.Builder{}

for y := range height {
	for x := range width {
		gradientContent.WriteString(
			lipgloss.NewStyle().
				Background(gradient2D[y][x]).
				Render(" "),
		)
	}
	if y < height-1 { // End of row.
		gradientContent.WriteString("\n")
	}
}
```

</details>

Thoughts? IMO, the former isn't that bad, so I don't think it's worth the extra resource overhead.

### To panic or not to panic

There is a handful of logic, specifically in the `Blend*` methods, around using various fallbacks if the user of the library provides potentially invalid data. Some examples:

- If some of the colors are invalid (e.g. nil), remove them from the color stops.
- If they provide a valid and invalid color, duplicate the valid color, so it won't be a gradient, but at least it won't panic.
- If `steps < 2`, always just output 2 steps, so in most cases, their `0` index will work, even if it isn't a gradient.

Typically, most of the lipgloss stuff doesn't return an error, which I suspect is because it's commonly used in `View()` methods which aren't supposed to produce actual logic, so returning an error probably doesn't make sense here. But what about panics for invalid input data? Fallback, or panic?

### DIFF
--- a/colors/colors.go
+++ b/colors/colors.go
@@ -1,0 +1,62 @@
+// Package colors provides a set of color-related utilities for working with colors.
+// This includes utilities for blending colors, adjusting brightness/alpha, etc.
+package colors
+
+import (
+	"cmp"
+	"image/color"
+
+	"github.com/lucasb-eyer/go-colorful"
+)
+
+func clamp[T cmp.Ordered](v, low, high T) T {
+	return min(high, max(low, v))
+}
+
+// ensureNotTransparent ensures that the alpha value of a color is not 0, and if
+// it is, we will set it to 1. This is useful for when we are converting from
+// RGB -> RGBA, and the alpha value is lost in the conversion for gradient purposes.
+func ensureNotTransparent(c color.Color) color.Color {
+	_, _, _, a := c.RGBA()
+	if a == 0 {
+		return Alpha(c, 1)
+	}
+	return c
+}
+
+// Alpha adjusts the alpha value of a color using a 0-1 (clamped) float scale
+// 0 = transparent, 1 = opaque.
+func Alpha(c color.Color, alpha float64) color.Color {
+	if c == nil {
+		return nil
+	}
+
+	r, g, b, _ := c.RGBA()
+	return color.RGBA{
+		R: uint8(min(255, float64(r>>8))),
+		G: uint8(min(255, float64(g>>8))),
+		B: uint8(min(255, float64(b>>8))),
+		A: uint8(clamp(alpha, 0, 1) * 255),
+	}
+}
+
+// Complementary returns the complementary color (180° away on color wheel) of
+// the given color. This is useful for creating a contrasting color.
+func Complementary(c color.Color) color.Color {
+	if c == nil {
+		return nil
+	}
+
+	// Offset hue by 180°.
+	cf, _ := colorful.MakeColor(ensureNotTransparent(c))
+
+	h, s, v := cf.Hsv()
+	h += 180
+	if h >= 360 {
+		h -= 360
+	} else if h < 0 {
+		h += 360
+	}
+
+	return colorful.Hsv(h, s, v).Clamped()
+}
--- a/colors/blending.go
+++ b/colors/blending.go
@@ -1,0 +1,194 @@
+package colors
+
+import (
+	"image/color"
+	"math"
+	"slices"
+
+	"github.com/lucasb-eyer/go-colorful"
+)
+
+// BlendLinear1D blends a series of colors together in one linear dimension using multiple
+// stops, into the provided number of steps. Uses the "CIE L*, a*, b*" (CIELAB) color-space.
+//
+// Note that if any of the provided colors are completely transparent, we will
+// assume that the alpha value was lost in conversion from RGB -> RGBA, and we
+// will set the alpha to opaque, as it's not possible to blend something completely
+// transparent.
+func BlendLinear1D(steps int, stops ...color.Color) []color.Color {
+	// Bound to a minimum of 2 steps. If they only provided one, it's actually invalid,
+	// but will ensure that we don't panic.
+	if steps < 2 {
+		steps = 2
+	}
+
+	// Ensure they didn't provide any nil colors.
+	stops = slices.DeleteFunc(stops, func(c color.Color) bool {
+		return c == nil
+	})
+
+	if len(stops) == 0 {
+		return nil // We can't safely fallback.
+	}
+
+	// If they only provided one valid color (or some nil colors), we will just return
+	// an array of that color, for the amount of steps they requested.
+	if len(stops) == 1 {
+		singleColor := stops[0]
+		result := make([]color.Color, steps)
+		for i := range result {
+			result[i] = singleColor
+		}
+		return result
+	}
+
+	blended := make([]color.Color, steps)
+
+	// Convert stops to colorful.Color once
+	cstops := make([]colorful.Color, len(stops))
+	for i, k := range stops {
+		cstops[i], _ = colorful.MakeColor(ensureNotTransparent(k))
+	}
+
+	numSegments := len(cstops) - 1
+	defaultSize := steps / numSegments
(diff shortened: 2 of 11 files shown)


lrstanley:
converting to draft, given the list of items I'll still need some feedback/opinions on under **Notes for reviewer**.

colors/colors.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	return colorful.Hsv(h, s, v).Clamped()
+}
+
+var errInvalidFormat = errors.New("invalid hex format")
+
+// FromHex parses a hex color string and returns a color.RGBA. The string can be
+// in the format #RRGGBB or #RGB. This is a more performant implementation of
+// [colorful.Hex].
```
aymanbagabas, line 69: Why don't you upstream this implementation in [go-colorful](https://github.com/lucasb-eyer/go-colorful)?

lrstanley: If you'd like me to upstream it, I can do that. Though:
- They specifically do different logic for float64 conversion and the rest of their library operates based on those converted values, rather than the minimal requirements we have (`color.Color` impl), so it may not see the same performance benefits.
- Ideally, the less we can rely on an external library, the better, especially for an implementation like this which I expect never needing to change (and it being relatively simple) -- and thus very low future maintenance.
  - I know there are still 2 other references to go-colorful, but those do rather complex color computation that I don't think would be maintainable to keep internal to the package.

aymanbagabas (commented):
This is great @lrstanley! Since we already have coloring functionality in the root package `lipgloss`, I wonder if it makes more sense to move these functions there 🤔 
We could also combine `FromHex` and the original `Color` into one.

lrstanley:
> > My gut says to move the color functionality to the top level, unless we want to move all existing color to a subpackage. @aymanbagabas, what do you think?
> 
> I think moving the color functionality to the top level is more idiomatic Go than moving them to a subpackage. Alternatively, to have a clear separation and also be idiomatic, we could move all the colors functionality into a separate package outside Lip Gloss. But perhaps that's something we can think about in the future.
> 
> For now, I would say let's move them to the top level.

I've moved everything to the root package. I have also made `FromHex` internal (-> `parseHex`), since I don't think we need to expand the packages API, as most users should be using the `Color` method.

aymanbagabas:
> Any thoughts on the items I outlined in the reviewer notes?

IMO, we could rename `BlendLinear1D` and `BlendLinear2D` to `Blend` and `Blend2D` respectively. Which is shorter and less confusing.

CC/ @meowgorithm

meowgorithm:
> IMO, we could rename `BlendLinear1D` and `BlendLinear2D` to `Blend` and `Blend2D` respectively. Which is shorter and less confusing.

Agreed, I'd prefer simpler names. Given that, how would we name radial gradients?

meowgorithm:
I'm down with that, actually as the discoverability makes sense. On that note we could also alias `Blend1D` to `BlendLinear1D` and so on for those who would prefer to be explicit.

lrstanley:
Anything else needed on this?

PR #559, opened 2025-08-23
Fix border width calculation in renderHorizontalEdge

akaoio:
## Summary
- Fixes an off-by-one error in `renderHorizontalEdge` that caused borders to render with incorrect width
- Ensures right border width is properly accounted for in total width calculation
- Corrects the loop logic to fill exactly the required middle section width

## Problem
The `renderHorizontalEdge` function had a bug where:
1. The right border width wasn't being properly subtracted when calculating the middle section
2. The loop iteration logic could cause borders to overflow or underflow their intended width

This was particularly noticeable with wider borders or specific width values.

## Solution
- Calculate middle width as `width - leftWidth - rightWidth`
- Fix the loop to iterate exactly over the middle width
- Add width accounting for right borders in `applyBorder`
- Add comprehensive tests to verify correct border rendering at various widths

## Test Plan
✅ Added `TestRenderHorizontalEdgeFix` with multiple test cases
✅ All existing tests pass
✅ Verified borders render at exact specified widths

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/borders.go
+++ b/borders.go
@@ -320,8 +320,11 @@ func (s Style) applyBorder(str string) string {
 		width += maxRuneWidth(border.Left)
 	}
 
-	if hasRight && border.Right == "" {
-		border.Right = " "
+	if hasRight {
+		if border.Right == "" {
+			border.Right = " "
+		}
+		width += maxRuneWidth(border.Right)
 	}
 
 	// If corners should be rendered but are set with the empty string, fill them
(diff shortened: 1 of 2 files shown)


meowgorithm:
Thanks for this. In order for us to verify the issue, can you put together some sample code to verify the issue exists? This will greatly help us expedite our ability to review the PR.

PR #560, opened 2025-08-24
feat(borders): initial border blend implementation

lrstanley:
## Changes

- Initial implementation of border blending (gradients).
  - New `Style.BorderBlendForeground(colors...)` method for storing the palette the user wants to use for blending.
  - New `Style.BorderBlendWrap(bool)` method for controlling how the palette is applied.
    - When wrapping is false (**the default**), your palette must specify an end-to-end gradient that ideally blends by itself (the start and end stitching well together).
    - When wrapping is true, the palette is used to generate half of the border, then the other half is the palette in reverse (A->B->C->B->A). This does make it simpler to create a blended border which has ends that stitch well together; however, it has the downside that when used with something like 3 colors, there is more of A and B than of C.
    - Does the name `Wrap` make sense for this?
  - Current blended border always starts from the same position (top left -> top right, -> bottom right -> bottom left -> top left).
  - Associated `Get<X>()` and `Unset<X>()` methods for the above 2 methods.

### Performance notes

- Naturally, blended borders have worse performance than regular borders, especially with the current implementation of how borders are implemented. I tried to keep the current implementation mostly the same for non-blended borders, though there is a small perf loss.
- I made some minor tweaks to other methods in lipgloss to help reduce some of the performance impact, and other simple things technically not related to borders (e.g. `Width()` should be 30-50% less CPU usage, lot of allocs -> 0 allocs).

## Checklist before requesting a review

- [x] I have read [`CONTRIBUTING.md`](https://github.com/charmbracelet/.github/blob/main/CONTRIBUTING.md)

---

## Reviewer Notes

- Only foreground blending is supported. I suspect it would be extremely unlikely that users would want background colors blended.
- The wrapping solution I use definitely makes things easier for users. Not sure if it should be the default (though it's not obvious that is how it would work, without maybe really good documentation), and/or if there is a better solution that I can't think of.
- Will still only render the specific border sides depending on what is enabled, but the blending is agnostic to it. As in, it will still internally generate a full blend of the full size, but only apply the colors to the visible borders.
- No current way to change the starting corner (or custom locations). I can easily add the ability to change which side it starts with if that's desired, but custom locations would be a challenge, unless we supported passing in a 1-to-1 colors array that exactly matches the number of border characters (easy for a user to mess up).
  - Maybe the base implementation should be "for simple use cases", and anything else, users could do themselves?
- Ideally, it would be nice to "cache" the generated blend to reduce compute on render, but I couldn't think of a simple way of doing this natively.
  - If you generate the blend during `set()` on the style, it won't correctly scale if the size fields change.
  - In one of my projects, I use an in-memory LFU cache based on the size, but that only works well when the palette/wrapping/etc is static.

### DIFF
--- a/borders.go
+++ b/borders.go
@@ -2,6 +2,7 @@ package lipgloss
 
 import (
 	"image/color"
+	"slices"
 	"strings"
 
 	"github.com/charmbracelet/x/ansi"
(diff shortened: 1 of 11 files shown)


lrstanley:
> In terms of API, what if we simply changed BorderForeground to be variadic so it can auto-upgrade to a blend?

Not opposed, though the only problem with variadic is that there is a potential that someone might provide 0 arguments. I assume in this situation, it's fine to have it be a no-op. As far as properties are concerned, we should probably still store them behind the scenes as separate properties to keep memory usage lower in the more common path.

> Does it make sense to group border options together in a functional argument kind of way? I don't think so, but throwing it out there as an option.

Given there wouldn't really be any other options that are structured in that way, will likely make the API feel a bit more weird. Especially because if we did, I'd think all other border options should also use a similar API, further segmenting the style API.

meowgorithm (approved):
Looks good to me. Let get this in.

PR #579, opened 2025-10-20
chore: rename `Complete` to `CompleteColor`

meowgorithm:
This revision renames `Complete` to `CompleteColor` and `CompleteFunc` to `CompleteColorFunc`. Out of context, `Complete` is too ambiguous, particularly because the scope of Lip Gloss is beyond color.

### DIFF
--- a/color.go
+++ b/color.go
@@ -232,7 +232,7 @@ func isDarkColor(c color.Color) bool {
 	return l < 0.5 //nolint:mnd
 }
 
-// CompleteFunc is a function that returns the appropriate color based on the
+// CompleteColorFunc is a function that returns the appropriate color based on the
 // given color profile.
 //
 // Example usage:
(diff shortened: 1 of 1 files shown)


aymanbagabas (commented):
Should we also rename `LightDark` to `LightDarkColor`? IMO, these two types don't really represent colors but more like color helpers.

lrstanley:
`CompleteColor` to me still doesn't convey the full purpose/intent, only helping convey that it's related to a color. Thoughts on something like `ResolveColor` (`ColorResolverFunc` for the func type)? "resolve" indicates it is the final result of some computed logic, and that might hint that a user can hook into that resolution logic to change it.

meowgorithm:
v2 is out: closing accordingly. Thanks everyone!

PR #591, opened 2025-11-20
v2: refactor: new Canvas, Compositor, and Layer API

aymanbagabas:
This change redefines what a Canvas and Layer are within the lipgloss
package. The Canvas is now a simple screen buffer that can compose
layers and drawable types.

A layer now doesn't do any computation and only represents a tree node. Using a compositor flattens a list of layers, and is able to draw them or render them into a string.

### DIFF
--- a/examples/go.mod
+++ b/examples/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/anmitsu/go-shlex v0.0.0-20200514113438-38f4b401e2be // indirect
 	github.com/charmbracelet/keygen v0.5.1 // indirect
 	github.com/charmbracelet/log/v2 v2.0.0-20251106192421-eb64aaa963a0 // indirect
-	github.com/charmbracelet/ultraviolet v0.0.0-20251201184010-2006a959db8b // indirect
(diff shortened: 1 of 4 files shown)


lrstanley (changes requested):
Some minor tweaks that may be worth looking at:

1. remove absolute Z functionality, as I don't think this will work how people expect it would.
   - This was causing the following to be rendered behind other lower z layers: https://cdn.liam.sh/share/2025/11/layers.txt (last layer in this go-spew output should be on top)
   - see also: https://cdn.liam.sh/share/2025/11/explorer_GQ10NjdU0j.mp4
   - maybe there is still a usecase for absolute z that I can't think of, so maybe it should be configurable on the compositor?
     - E.g. `(c *Compositor) UseAbsoluteZ(bool)`. Alternatively, on `*Layer`, we could also do `(l *Layer) CompoundingZ(int)`, which sets both a `z` field, in addition to an `compoundingZ` field, and if that is present, it's always added onto children?
2. Allow `lipgloss.Canvas` to be reused, for users who would like higher perf.

Patch for the above: https://cdn.liam.sh/share/2025/11/misc-v2-canvas-1.patch

```console
$ curl -sL https://cdn.liam.sh/share/2025/11/misc-v2-canvas-1.patch | git apply
```

PR #641, opened 2026-03-31
fix: typo in table height comment

Rohan5commit:
## Summary
- Fix the possessive typo in the `computeHeight` comment.

## Related issue
- N/A

## Guideline alignment
- Comment-only change in one file.
- No behavior changes.

## Validation
- Not run locally; this is a comment-only change.

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -367,7 +367,7 @@ func (t *Table) String() string {
 		Render(strings.TrimSuffix(sb.String(), "\n"))
 }
 
-// computeHeight computes the height of the table in it's current configuration.
+// computeHeight computes the height of the table in its current configuration.
 func (t *Table) computeHeight() int {
 	hasHeaders := len(t.headers) > 0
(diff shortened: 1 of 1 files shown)


MukundaKatta:
Nice catch on the grammar — worth noting the same possessive-vs-contraction slip is present in five more doc comments in `table/table.go`:

- L109: \`StyleFunc sets the style for a cell based on it's position\` → `its`
- L115: \`style returns the style for a cell based on it's position\` → `its`
- L383: \`constructs the top border for the table given it's current\` → `its`
- L402: \`constructs the bottom border for the table given it's current\` → `its`
- L421: \`constructs the headers for the table given it's current\` → `its`

(L320 \`until it's the same length\` and the two `it's` uses in `style.go` are legitimate contractions of "it is", so those should stay as-is.)

Up to you whether to roll them into this PR or leave a follow-up — the scope expands from 1→6 lines of the same file either way.

PR #702, opened 2026-06-20
fix: pad horizontal border edge with wide middle runes

s3onghyun:
## Summary

A custom border whose `Top`/`Bottom` is a **wide (2-cell) rune** (e.g. CJK or box-art) renders the box one cell too wide when the inner span is odd. The top/bottom edges then don't line up with the body:

```
before:            after:
┌ああ┐  (w=6)       ┌あ ┐  (w=5)
│abc│  (w=5)        │abc│  (w=5)
└ああ┘  (w=6)       └あ ┘  (w=5)
```

## Cause

In `renderHorizontalEdge` (`borders.go`) the fill loop checks `i < inner` **before** writing, then advances `i` by the rune's full cell width. When the middle rune is 2 cells wide and only one cell of space remains, it still writes the whole rune, overshooting the target width by one.

## Fix

If the next rune is too wide for the remaining space, write a single space instead so the edge lands exactly on the requested width. Narrow (1-cell) runes and even inner widths are unaffected — the built-in borders all use 1-cell runes, so default rendering doesn't change.

## Test

`TestRenderHorizontalEdgeWideMiddle` asserts `ansi.StringWidth(out) == width` across wide/narrow runes and odd/even inner spans. It fails before the change (`width = 8, want 7`) and passes after. `go test ./...` is green.

### DIFF
--- a/borders.go
+++ b/borders.go
@@ -509,10 +509,21 @@ func renderHorizontalEdge(left, middle, right string, width int) string {
 	out := strings.Builder{}
 	out.WriteString(left)
 
-	for i := 0; i < width-leftWidth-rightWidth; {
+	inner := width - leftWidth - rightWidth
+	for i := 0; i < inner; {
 		r := runes[j]
+		rw := ansi.StringWidth(string(r))
+		if i+rw > inner {
+			// The next rune is too wide to fit in the remaining space (e.g. a
+			// wide middle rune against an odd inner width). Pad with a single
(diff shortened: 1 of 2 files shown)


s3onghyun:
Closing this to refocus my open-source time on a smaller set of projects. The change still stands on its own — happy to reopen if a maintainer would like to pick it up. Thanks for maintaining this!

PR #704, opened 2026-06-22
fix(tree): remove incorrect start/end swap in Offset

### LINKED ISSUE #535: Tree Offset logic is broken
**Describe the bug**
Offset always excludes more rows from the end as compared with the start of the list

**Setup**

**To Reproduce**
in the [duckduckgoose](https://github.com/charmbracelet/lipgloss/tree/master/examples/list/duckduckgoose) example add following line after line 24
	l.Offset(2, 1)

**Current behavior**
  Duck
  Duck

**Expected behavior**
       Duck
Honk → Goose

**Solution**
The code for swapping the start and end values if start > end should be removed from the Tree Offset function

xbrxr03:
## Problem

`Tree.Offset(start, end)` incorrectly swaps `start` and `end` when `start > end`. This makes `Offset(2, 1)` behave like `Offset(1, 2)`, hiding rows from the wrong end of the tree.

Reported in #535 — calling `Offset(2, 1)` on `[Duck, Duck, Goose, Duck, Duck]` should skip 2 from the start and 1 from the end, but the swap inverts the meaning so you end up skipping 1 from the start and 2 from the end instead.

## Fix

- **Removed the start/end swap** in `Offset()`. The parameters now mean what they say: `start` = rows to skip from the beginning, `end` = rows to skip from the end. No magic reordering.
- **Added bounds clamping in `Children()`** so out-of-range offsets produce empty results instead of panicking on negative loop bounds.
- **Negative values are clamped to zero** in `Offset()`.
- **Improved doc comment** on `Offset()` to explain what the parameters actually do.

## Comparison with #612

This PR builds on the same core insight as #612 (remove the swap), but goes further:

1. **Bounds safety**: The existing `Children()` method could produce negative loop bounds if `offset[1] > children.Length()`. This PR adds explicit clamping so the iteration is always safe.
2. **Better negative handling**: Old code treated negative `end` as "use children length" — this treated `Offset(0, -1)` as "show everything". The new behavior treats all negative values as zero, which is more intuitive.
3. **Comprehensive tests**: 9 test cases covering zero offset, skip from start, skip from end, skip from both, `start > end` (the core bug), offset larger than children, negative values, offset that swallows all items, and equal offsets.

## Test Results

```
=== RUN   TestTreeOffset
--- PASS: TestTreeOffset (0.00s)
    --- PASS: zero_offset
    --- PASS: skip_from_start
    --- PASS: skip_from_end
    --- PASS: skip_from_both
    --- PASS: start_greater_than_end
    --- PASS: offset_larger_than_children
    --- PASS: negative_values_clamped_to_zero
    --- PASS: offset_swallows_all
    --- PASS: equal_offset
```

Full suite passes: `ok  charm.land/lipgloss/v2/tree  0.216s`

Closes #535

### DIFF
--- a/tree/tree.go
+++ b/tree/tree.go
@@ -118,19 +118,17 @@ func (t *Tree) Hide(hide bool) *Tree {
 // SetHidden hides a Tree node.
 func (t *Tree) SetHidden(hidden bool) { t.Hide(hidden) }
 
-// Offset sets the Tree children offsets.
+// Offset sets the Tree children offsets. Start is the number of children to
+// skip from the beginning, and end is the number of children to skip from the
+// end. For example, Offset(2, 1) on [A, B, C, D, E] yields [C, D].
+//
+// Passing a negative value for either parameter treats it as zero (no offset).
 func (t *Tree) Offset(start, end int) *Tree {
-	if start > end {
-		_start := start
-		start = end
-		end = _start
-	}
-
 	if start < 0 {
 		start = 0
 	}
-	if end < 0 || end > t.children.Length() {
-		end = t.children.Length()
+	if end < 0 {
+		end = 0
 	}
 
 	t.offset[0] = start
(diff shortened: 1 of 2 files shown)


xbrxr03:
Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

PR #705, opened 2026-06-22
feat(table): add OverflowRow option to disable overflow row

### LINKED ISSUE #557: v2: Add option to disable table overflow row
**Is your feature request related to a problem? Please describe.**
I don't want an overflow row on a table. I'm not sure what it's original intended purpose was, presumably it makes sense when the user is being informed that there are more rows than their terminal window size permits and that they need to make their terminal window bigger to accomodate the extra rows?

But I'm employing pagination and in that scenario, an overflow row is arguably superfluous.

**Describe the solution you'd like**
Add an option to disable the default overflow row.

xbrxr03:
## Problem

When a table has more data rows than can fit in its configured `Height`, lipgloss renders an overflow row with `…` in each column. This tells the user there are more rows below. But if you're implementing pagination, that overflow row is redundant — you already know there are more pages.

Reported in #557.

## Fix

Added `OverflowRow(bool)` and `GetOverflowRow()` to `Table`. Default is `true` (preserve existing behavior). Set to `false` to suppress the overflow row entirely.

```go
table := table.New().
    Headers("Name", "Country").
    Row("Alice", "US").
    Row("Bob", "UK").
    Row("Carol", "DE").
    Row("Dave", "FR").
    Row("Eve", "JP").
    Height(5).
    OverflowRow(false)   // <-- no more "…" row
```

The change is minimal: a `bool` field on `Table`, a setter, a getter, and one condition in `String()` that gates the overflow row rendering. No changes to the height/scroll calculation logic.

## Tests

4 new sub-tests under `TestOverflowRowDisabled`:
- `overflow_enabled_default`: confirms `…` appears with default settings
- `overflow_disabled`: confirms `…` is gone when `OverflowRow(false)`
- `overflow_disabled_shows_all`: confirms output is identical when all rows fit regardless of setting
- `getter`: confirms `GetOverflowRow()` returns correct defaults and after mutation

Full test suite: `ok  charm.land/lipgloss/v2/table  0.279s`

Closes #557

### DIFF
--- a/table/table.go
+++ b/table/table.go
@@ -71,6 +71,7 @@ type Table struct {
 	firstVisibleRowIndex int
 	lastVisibleRowIndex  int
 	overflowHeight       int
+	overflowRow          bool
 }
 
 // New returns a new Table that can be modified through different
@@ -88,6 +89,7 @@ func New() *Table {
 		borderRight:  true,
 		borderTop:    true,
 		wrap:         true,
+		overflowRow:  true,
 		data:         NewStringData(),
 	}
 }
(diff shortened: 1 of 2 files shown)


xbrxr03:
Closing this PR due to lack of maintainer engagement. I'm happy to reopen if there's interest. Thanks for the great project.

Issue #745, opened 2026-09-19
Windows: HasDarkBackground blocks forever when stdin is not a console (CONIN$ fallback is not cancellable)


keppy:
**What happens**

On Windows, `lipgloss.HasDarkBackground` (and `BackgroundColor`) never returns when `os.Stdin` is not a console. The query's own 2-second timeout does not fire, so the call blocks for the life of the process.

**Why**

`BackgroundColor`'s Windows branch substitutes the console whenever either stream is not a terminal:

```go
if !term.IsTerminal(in.Fd()) {
    f, err := os.OpenFile("CONIN$", os.O_RDWR, 0o644)
    if err != nil { ... }
    in = f
}
```

`queryTerminal` then builds its cancelable reader with `uv.NewCancelReader(in)`. ultraviolet returns its cancellable `conInputReader` only when the file *is* `os.Stdin` and `GetConsoleMode` succeeds:

```go
if f, ok := r.(cancelreader.File); !ok || f.Fd() != os.Stdin.Fd() ||
    windows.GetConsoleMode(windows.Handle(f.Fd()), &dummy) != nil {
    return fallback(r)   // cancelreader.fallbackCancelReader
}
```

Handed the freshly opened CONIN$, it takes the fallback. That reader's own doc comment says it "does not actually support cancelation during an ongoing Read() call. Thus, Cancel() always returns false" — it only makes *subsequent* reads return `ErrCanceled`. The timeout goroutine therefore fires into a no-op and the read stays parked in `syscall.ReadConsole` forever.

#635 is the same bug on the Unix side and was fixed by #636, which added

```go
if !term.IsTerminal(in.Fd()) || !term.IsTerminal(out.Fd()) {
    return nil, fmt.Errorf("input/output is not a terminal")
}
```

— but that guard sits *after* the Windows branch's `return backgroundColor(in, out)`, so Windows still substitutes CONIN$ and blocks. (Likely related to #73 as well.)

**Repro**

```go
package main

import (
	"fmt"
	"os"

	lipgloss "charm.land/lipgloss/v2"
)

func main() {
	fmt.Println(lipgloss.HasDarkBackground(os.Stdin, os.Stdout))
}
```

On Windows 11, from a session that has a console, with stdin redirected:

```
$ ./probe < /dev/null     # never returns
```

No TTY needed, no CI needed. The console must exist for `CONIN$` to open — a process with no console takes the error path instead and is unaffected, which is why this hides on some machines.

[...]

PR #746, opened 2026-09-22
fix: avoid Windows background query hangs for non-terminals

sb123sb123:
Issue: https://github.com/charmbracelet/lipgloss/issues/745

## Problem

On Windows, \`BackgroundColor\` replaces non-terminal input/output with \`CONIN$\`/\`CONOUT$\`. The resulting reader cannot be canceled by the terminal query timeout, so a call with redirected or piped stdin can block forever.

Before this change, running the issue reproduction with stdin redirected (for example, \`probe.exe < NUL\`) can hang indefinitely while a console is available.

## Fix

Reject non-terminal input or output before entering the Windows console fallback, matching the existing Unix guard. The call now returns \`input/output is not a terminal\`; \`HasDarkBackground\` can then use its documented error fallback. A regression test covers non-terminal files.

After this change, the same redirected-input call returns promptly instead of hanging.

## Tests

- \`go test -run '^TestBackgroundColorRejectsNonTerminal$' -count=1 .\` — passed on Windows/amd64 with Go 1.26.7.
- \`go test -count=1 ./...\` — the root package passed, but subpackage tests could not complete in this environment because \`github.com/aymanbagabas/go-udiff v0.4.1\` and \`github.com/charmbracelet/x/exp/golden\` were not present in the module cache and the Go proxy was unreachable.

AI assistance: OpenAI Codex (GPT-5.6 Luna)


keppy:
there goes my hero

keppy:
sb123sb123

